PR #247, opened 2023-01-19
format links in chat messages

paulwe:
following up on the suggestions from @Ocupe in https://github.com/livekit/components-js/pull/225 and the discussion with @lukasIO the formatter shouldn't affect the component bundle unless it's explicitly requested. i'll add this to meet once this is landed.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -39,7 +39,9 @@
     "@floating-ui/react": "^0.15.1",
     "@livekit/components-core": "0.1.13",
     "@react-hook/resize-observer": "^1.2.6",
-    "clsx": "^1.2.1"
+    "clsx": "^1.2.1",
+    "email-regex": "^5.0.0",
+    "url-regex": "^5.0.0"
   },
   "peerDependencies": {
     "react": ">=18",
@@ -54,8 +56,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "tsup": "^6.5.0",
-    "typescript": "^4.9.4",
-    "url-regex": "^5.0.0"
+    "typescript": "^4.9.4"
   },
   "engines": {
     "node": ">=10"
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -54,7 +54,8 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "tsup": "^6.5.0",
-    "typescript": "^4.9.4"
+    "typescript": "^4.9.4",
+    "url-regex": "^5.0.0"
   },
   "engines": {
     "node": ">=10"
--- a/packages/react/src/prefabs/VideoConference.tsx
+++ b/packages/react/src/prefabs/VideoConference.tsx
@@ -9,8 +9,11 @@ import { TileLoop } from '../components/TileLoop';
 import { Chat } from './Chat';
 import { ConnectionStateToast } from '../components/Toast';
 import { useMediaQuery } from '../hooks/utiltity-hooks';
+import { MessageFormatter } from './ChatEntry';
 
-export type VideoConferenceProps = React.HTMLAttributes<HTMLDivElement>;
+export interface VideoConferenceProps extends React.HTMLAttributes<HTMLDivElement> {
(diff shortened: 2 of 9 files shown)


packages/react/src/utils.ts
```diff
@@ -73,4 +74,35 @@ const useSize = (target: React.RefObject<HTMLDivElement>) => {
...
+    })
+    .filter(({ index }, i, arr) => {
+      if (i === 0) return true;
+      const prev = arr[i - 1];
+      return prev.index + prev.content.length <= index;
+    });
+
+  const tokens = [];
+  let pos = 0;
+  for (const { type, content, index } of matches) {
```
lukasIO, line 102: this is currently only looping over matches, without special handling for the case when there's no match at all. This results in messages that are a string without any matches to be returned as an empty array

packages/react/package.json
```diff
@@ -49,12 +49,14 @@
     "@svgr/cli": "^6.5.1",
     "@types/react": "^18.0.25",
     "@types/react-dom": "^18.0.8",
+    "email-regex": "^5.0.0",
```
lukasIO, line 52: the regex packages should probably go into "dependencies" instead of "devDependencies"

packages/react/src/prefabs/VideoConference.tsx
```diff
@@ -9,8 +9,11 @@ import { TileLoop } from '../components/TileLoop';
 import { Chat } from './Chat';
 import { ConnectionStateToast } from '../components/Toast';
 import { useMediaQuery } from '../hooks/utiltity-hooks';
+import { MessageFormatter } from './ChatEntry';
 
-export type VideoConferenceProps = React.HTMLAttributes<HTMLDivElement>;
+export interface VideoConferenceProps extends React.HTMLAttributes<HTMLDivElement> {
+  formatChatMessage?: MessageFormatter;
```
Ocupe, line 15: `formatChatMessage` could also be the name of a boealean variable. What do you think of `chatMessageFormatter` to avoid ambiguity?

packages/react/src/prefabs/VideoConference.tsx
```diff
@@ -53,7 +56,10 @@ export function VideoConference({ ...props }: VideoConferenceProps) {
           )}
           <ControlBar variation={isMobile ? 'minimal' : 'verbose'} controls={{ chat: true }} />
         </div>
-        <Chat style={{ display: widgetState.showChat ? 'flex' : 'none' }} />
+        <Chat
+          style={{ display: widgetState.showChat ? 'flex' : 'none' }}
+          formatMessage={formatChatMessage}
```
Ocupe, line 61: Same here: `formatMessage` could maybe be renamed to `messageFormater`?

packages/react/src/utils.ts
```diff
@@ -73,4 +74,35 @@ const useSize = (target: React.RefObject<HTMLDivElement>) => {
   return size;
 };
 
-export { mergeProps, LKComponentAttributes, useObservableState, cloneSingleChild, useSize };
+export type TokenizeGrammar = { [type: string]: RegExp };
+
+export function tokenize(input: string, grammar: TokenizeGrammar) {
+  const matches = Object.entries(grammar)
```
Ocupe, line 80: Not important: `type TokenizeGrammer` and `function tokenize` could probably be moved to `/packages/core/` as they are not React specific and could be used in other implementations in the future.

Ocupe (approved):
Looking good! Only optional little things to complain about. Thank you for the PR :)

packages/react/src/utils.ts
lukasIO: ah, sorry paul, that's my fault. @Ocupe is right, I misdirected you to `utils` in the react package for everything.

lukasIO:
cc @mdo should links within chat messages be visually different from the rest of the text?

packages/react/src/utils.ts
paulwe: good catch

lukasIO (approved):
nice! LGTM, we can always add additional styles for the links later on!

PR #255, opened 2023-01-22
Use explicit aspect ratio in focus-aside

lukasIO:
also fixes the `yarn link` bug for the css package.

### DIFF
--- a/packages/styles/package.json
+++ b/packages/styles/package.json
@@ -19,7 +19,7 @@
       "types": "./dist/types_unprefixed/styles.scss.d.ts"
     }
   },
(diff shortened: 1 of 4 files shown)


packages/styles/package.json
```diff
@@ -19,7 +19,7 @@
       "types": "./dist/types_unprefixed/styles.scss.d.ts"
     }
   },
-  "main": "dist/styles.css",
+  "main": "dist/general/styles.css",
```
Ocupe, line 22: 💯

Ocupe (approved):

PR #258, opened 2023-01-23
Jonas/lk 2323 handle data attribute edge case in

Ocupe:
- [x] add tests for PostCSS plugin
- [x] handle edge case: use of data attributes in `attr()` function like `attr(data-participant-name)`


lukasIO (approved):
nice!

PR #266, opened 2023-01-25
Floating UI revision

Ocupe:
- [x] replace floating-ui/react with floating-ui/dom
- [x] fix buggy positioning of MediaDeviceMenu
- [x] handle updates of media device list while menu is open


lukasIO (approved):
nice!

PR #273, opened 2023-01-26
update floating ui on resize

Ocupe:


packages/react/src/prefabs/MediaDeviceMenu.tsx
```diff
@@ -68,10 +70,12 @@ export const MediaDeviceMenu = ({
 
   React.useEffect(() => {
     document.addEventListener<'click'>('click', handleClickOutside);
+    window.addEventListener<'resize'>('resize', () => setUpdateRequired(true));
     return () => {
       document.removeEventListener<'click'>('click', handleClickOutside);
+      window.removeEventListener<'resize'>('resize', () => setUpdateRequired(true));
```
lukasIO, line 76: resize handling is something we will have to have across all frameworks too right? 
maybe moving it into core makes sense?

PR #274, opened 2023-01-26
Add import assertion for json import

lukasIO:
The lack of the assertion was causing crashes on nodeJS versions > 17.5 due to import assertions being mandatory https://nodejs.org/api/esm.html#import-assertions

First attempt to fix was to just add the `assert {type: 'json'}`. In order to maximise support across nodejs environments for the packages I then just replaced the dependency with a different one that's not using JSON as an export format.


Ocupe (approved):
👍🏼

PR #277, opened 2023-01-27
Add High level data channel APIs

lukasIO:
This generalises usage for data channels. 
One example where the generalised approach is used can be seen in the chat.ts in the core package.


lukasIO:
@Ocupe will merge for now, but would be great to get your feedback on this!

PR #295, opened 2023-02-06
Add useRemoteParticipant hook

lukasIO:


packages/react/src/hooks/participant-hooks.ts
```diff
@@ -86,21 +92,31 @@ export const useLocalParticipant = () => {
...
+};
+
 /**
  * The useRemoteParticipants
  */
 export const useRemoteParticipants = ({
   filter,
 }: {
-  filter?: (participants: Array<Participant>) => Array<Participant>;
+  filter?: (participant: Participant) => boolean;
```
Ocupe, line 111: Should we use the `RemoteParticipants` type here?
To be consistent with the other filters, I would also suggest adding a filter type to `core` and using it here. Like: https://github.com/livekit/components-js/blob/270c8f5a79a50e09f4846bee81fbefc959d3b02c/packages/core/src/types.ts#L21-L22

Ocupe (approved):

packages/react/src/hooks/participant-hooks.ts
lukasIO: yeah, sorry, this was just me being lazy, will address it.

PR #298, opened 2023-02-08
Use consistent parameters style for hooks

lukasIO:


Ocupe (approved):
Looking good! 👍 Two little notes:

1. I'm not a big fan of abbreviations for readability reasons, and with editor autocomplete the penalty for having long names is pretty low. On the other hand, I'm not a big fan of long names either, so I understand why you did it. Considering both sides, I would still vote for something like `UseTrackMutedIndicatorOptions` over `UseTrackMutedIndicatorOpts`.
2. In the [Deno style guide](https://deno.land/manual@v1.30.2/references/contributing/style_guide#exported-functions-max-2-args-put-the-rest-into-an-options-object) the options are never optional, but the functions that use them give a default value of an empty object. Is this intentional because of the React default option problem? As a personal goal, I try to reduce and remove as many optional variables from the code as possible. Just to keep the cardinality of the variables/parameters as low as possible and thus reduce the number of if-not-null checks.

lukasIO:
both great points, thanks! addressed them :)

PR #304, opened 2023-02-10
Replace pin reducer with pin observable

lukasIO:


Ocupe:
@lukasIO this is currently failing for me with:
```
@livekit/component-example-next:build: ./pages/huddle.tsx:124:11
@livekit/component-example-next:build: Type error: Property 'state' does not exist on type 'PinContextType'.
@livekit/component-example-next:build: 
@livekit/component-example-next:build:   122 | 
@livekit/component-example-next:build:   123 | const CustomFocus = () => {
@livekit/component-example-next:build: > 124 |   const { state } = useLayoutContext().pin;
```

PR #306, opened 2023-02-13
ParticipantLoop updateOn proposal 

Ocupe:
- [x] Group RoomEvents into nice logical units
  - [x] Check grouped events and naming for grouping 
- [x] Double check naming
  - [x] updateOnlyOn
  - [x] RoomEventGroup

### DIFF
--- a/packages/core/src/helper/eventGroups.ts
+++ b/packages/core/src/helper/eventGroups.ts
@@ -2,6 +2,18 @@ import { RoomEvent } from 'livekit-client';
 
 export type UpdateOnlyOn = (RoomEvent[] | RoomEvent)[];
 
+export function toRoomEvents(updateOnlyOn: UpdateOnlyOn): RoomEvent[] {
+  const events: RoomEvent[] = [];
+  updateOnlyOn.forEach((value) => {
+    if (Array.isArray(value)) {
+      events.concat(value);
+    } else {
+      events.push(value);
+    }
+  });
+  return events;
+}
+
 /**
  * RoomEventGroup is a helper object of RoomEvents that are logically grouped together.
  * The intention is to provide a high-level way to get all the RoomEvents needed to subscribe to certain state changes.
--- a/packages/core/src/helper/eventGroups.ts
+++ b/packages/core/src/helper/eventGroups.ts
@@ -1,9 +1,21 @@
 import { RoomEvent } from 'livekit-client';
 
-export const RoomEventGroup = {
-  /** Subscribe to track updates. */
+export type UpdateOnlyOn = (RoomEvent[] | RoomEvent)[];
+
+/**
+ * RoomEventGroup is a helper object of RoomEvents that are logically grouped together.
+ * The intention is to provide a high-level way to get all the RoomEvents needed to subscribe to certain state changes.
+ *
+ * @example
+ * ```tsx
+ *   // Get all events needed to listen to tack updates.
+ *   const events = RoomEventGroup.trackUpdates
+ * ```
+ */
+export const RoomEventGroup: Record<string, RoomEvent[]> = {
+  /**  */
   identity: [RoomEvent.ParticipantMetadataChanged],
-  /** Subscribe to track updates. */
+  /** Update on participant track updates. */
   trackUpdates: [
     RoomEvent.TrackMuted,
     RoomEvent.TrackUnmuted,
@@ -12,7 +24,7 @@ export const RoomEventGroup = {
     RoomEvent.TrackSubscribed,
     RoomEvent.TrackUnsubscribed,
   ],
-  /** Collection of events around participant connection. */
+  /** Update on participants joining and leaving the room. */
   participant: [
     RoomEvent.ParticipantConnected,
     RoomEvent.ParticipantDisconnected,
@@ -22,4 +34,4 @@ export const RoomEventGroup = {
   metadata: [RoomEvent.ParticipantMetadataChanged],
   /** Collection of all RoomEvents.  */
   all: [],
-} as const;
+};
--- a/packages/core/src/helper/eventGroups.ts
+++ b/packages/core/src/helper/eventGroups.ts
@@ -1,60 +1,23 @@
 import { RoomEvent } from 'livekit-client';
 
-export type UpdateOnlyOn = (RoomEvent[] | RoomEvent)[];
+export const allRemoteParticipantRoomEvents = [
+  RoomEvent.ConnectionStateChanged,
+  RoomEvent.RoomMetadataChanged,
 
-export function toRoomEvents(updateOnlyOn: UpdateOnlyOn): RoomEvent[] {
-  const events: RoomEvent[] = [];
-  updateOnlyOn.forEach((value) => {
-    if (Array.isArray(value)) {
-      events.concat(value);
-    } else {
-      events.push(value);
-    }
-  });
-  return events;
-}
+  RoomEvent.ActiveSpeakersChanged,
+  RoomEvent.ConnectionQualityChanged,
 
-/**
- * RoomEventGroup is a helper object of RoomEvents that are logically grouped together.
- * The intention is to provide a high-level way to get all the RoomEvents needed to subscribe to certain state changes.
- *
- * @example
- * ```tsx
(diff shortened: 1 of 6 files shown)


packages/react/src/hooks/participant-hooks.ts
```diff
@@ -14,38 +15,36 @@ import {
...
+ * By default the hook updates on all relevant RoomEvents to keep the returned participants array up to date.
  */
 export const useParticipants = (options: UseParticipantsOptions = {}) => {
   const [participants, setParticipants] = React.useState<Participant[]>([]);
-  const remoteParticipants = useRemoteParticipants({ filter: undefined });
+  const { updateOnlyOn } = options;
+  const remoteParticipants = useRemoteParticipants({
+    updateOnlyOn: updateOnlyOn ? toRoomEvents(updateOnlyOn) : [],
+  });
   const { localParticipant } = useLocalParticipant();
```
lukasIO, line 43: should we forward the update restriction also to the local participant?

packages/react/src/hooks/participant-hooks.ts
```diff
@@ -14,38 +15,36 @@ import {
...
-  const filterDependencies = React.useMemo(
-    () => options?.filterDependencies ?? [],
-    [options?.filterDependencies],
-  );
 
   React.useEffect(() => {
```
lukasIO, line 45: can we maybe even get rid of the useEffect here without causing constant re-renders?

packages/react/src/hooks/participant-hooks.ts
```diff
@@ -14,38 +15,36 @@ import {
...
+  UpdateOnlyOn,
+  toRoomEvents,
 } from '@livekit/components-core';
 import { useEnsureParticipant, useRoomContext } from '../context';
 import { useObservableState } from '../helper/useObservableState';
 
 export interface UseParticipantsOptions {
-  filter?: ParticipantFilter;
-  filterDependencies?: Array<unknown>;
+  updateOnlyOn?: UpdateOnlyOn;
```
lukasIO, line 27: This type doesn't make it clear, that in most cases you would want to use the grouped room events

lukasIO: it suggests passing in room events directly

packages/react/src/hooks/participant-hooks.ts
```diff
@@ -116,20 +116,23 @@ export interface UseRemoteParticipantsOptions {
 export const useRemoteParticipants = (options: UseRemoteParticipantsOptions = {}) => {
   const room = useRoomContext();
   const [participants, setParticipants] = React.useState<RemoteParticipant[]>([]);
+  const { filter, updateOnlyOn } = options;
```
lukasIO, line 119: I think we would want to get rid of `filter` here

packages/react/src/hooks/participant-hooks.ts
```diff
@@ -116,20 +116,23 @@ export interface UseRemoteParticipantsOptions {
...
       }
       setParticipants(participants);
     },
-    [options.filter],
+    [filter],
   );
   React.useEffect(() => {
-    const listener = connectedParticipantsObserver(room).subscribe(handleUpdate);
+    const listener = connectedParticipantsObserver(room, {
+      extraRoomEvents: updateOnlyOn ?? [],
```
lukasIO, line 132: shouldn't this default to `all` instead?

packages/core/src/helper/eventGroups.ts
```diff
@@ -0,0 +1,60 @@
...
+ * RoomEventGroup is a helper object of RoomEvents that are logically grouped together.
+ * The intention is to provide a high-level way to get all the RoomEvents needed to subscribe to certain state changes.
+ *
+ * @example
+ * ```tsx
+ *   // Get all events needed to listen to tack updates.
+ *   const events = RoomEventGroup.trackUpdates
+ * ```
+ */
+export const RoomEventGroup: Record<string, RoomEvent[]> = {
```
lukasIO, line 27: I think we'll need to do a pass on naming. 
both for the "group" and also for its members. 

I'd be in favor of naming it around the Participants from the start, as - at leaast in this case - the use case is on the useParticipants hook.

packages/react/src/components/ParticipantLoop.tsx
```diff
@@ -36,9 +37,20 @@ type ParticipantLoopProps = {
...
+  updateOnlyOn,
   ...props
 }: React.PropsWithChildren<ParticipantLoopProps>) => {
-  const participants = useParticipants({ filter, filterDependencies });
+  const participants_ = useParticipants({ updateOnlyOn });
+  const filterDependenciesArray = React.useMemo(
+    () => filterDependencies ?? [],
+    [filterDependencies],
+  );
+  const participants = React.useMemo(() => {
```
lukasIO, line 48: just around naming:
```suggestion
const filteredParticipants = React.useMemo(() => {
```

packages/react/src/components/layout/GridLayout.tsx
```diff
@@ -26,7 +26,14 @@ export interface GridLayoutProps extends React.HTMLAttributes<HTMLDivElement> {
  * ```
  */
 export function GridLayout({ filter, filterDependencies, ...props }: GridLayoutProps) {
-  const participants = useParticipants({ filter, filterDependencies });
+  const participants_ = useParticipants();
+  const participants = React.useMemo(() => {
```
lukasIO, line 30: same as above, I'd rather see `filteredParticipants` than `participants_`

packages/react/src/components/layout/GridLayout.tsx
```diff
@@ -43,7 +50,7 @@ export function GridLayout({ filter, filterDependencies, ...props }: GridLayoutP
       gridEl.current.style.setProperty('--lk-col-count', cols.toString());
       gridEl.current.style.width = `${widthAdjust}%`;
     }
-  }, [width, height, participants, gridEl.current]);
+  }, [width, height, participants, gridEl]);
```
lukasIO, line 53: did you double check that this doesn't have re-render side effects?
I think that was the initial reason why I ignored the eslint warning here for `.current`

packages/core/src/observables/participant.ts
```diff
@@ -165,20 +166,34 @@ export function createIsSpeakingObserver(participant: Participant) {
   );
 }
 
-export function connectedParticipantsObserver(room: Room) {
+type ConnectedParticipantsObserverOptions = {
+  extraRoomEvents?: RoomEvent[];
```
lukasIO, line 170: In my thinking the events aren't "extra", but rather the actual one's we update on. 
E.g. they default to `all` room events and if you specify them, that default is overridden with whatever you pipe in.

lukasIO: maybe that would lead to `connectedParticipantsObserver` to become `participantsObserver` and the events always include the three main one's?

packages/react/src/hooks/participant-hooks.ts
Ocupe: Lets try! 👍

packages/react/src/hooks/participant-hooks.ts
Ocupe: Agreed. Do you have a name in mind?

packages/core/src/helper/eventGroups.ts
Ocupe: I'm not a fan of the names either. I'm happy to hear your suggestions.

packages/react/src/components/layout/GridLayout.tsx
Ocupe: I didn't -> reverted 👍

packages/core/src/observables/participant.ts
Ocupe: What you describe sounds like we have three main events that we always want to include, even if the user decides to override the default (all events). 
If we have a base and some additional (user-defined) events, I find my naming pretty descriptive. Don't you think?

packages/react/src/hooks/participant-hooks.ts
lukasIO: decided on no

PR #308, opened 2023-02-14
use updateOnlyOn logic for remote participant hook

lukasIO:


packages/core/src/observables/participant.ts
```diff
@@ -201,13 +201,32 @@ export function connectedParticipantsObserver(
...
+) {
+  const additionalEvents = options.additionalEvents ?? allRemoteParticipantEvents;
   const observable = observeRoomEvents(
     room,
     RoomEvent.ParticipantConnected,
     RoomEvent.ParticipantDisconnected,
     RoomEvent.ConnectionStateChanged,
-  ).pipe(map((r) => r.getParticipantByIdentity(identity) as RemoteParticipant | undefined));
+  ).pipe(
+    mergeMap((r) => {
```
Ocupe, line 220: I was wondering if we should use a `switchMap` (https://rxjs.dev/api/operators/switchMap) operator instead. It's a drop in replacement that eliminates the risk of having multiple inner observables emitting.
Another thought. Right now we run `observeParticipantEvents` on every new room event, right? Maybe we could use `switchScan` to keep track of the observable returned by `observeParticipantEvents`?

Ocupe (commented):
Nice find with the `mergeMap`! 👍

PR #314, opened 2023-02-15
Limit styles to components

lukasIO:
reasoning being that in cases where people integrate the LK room in their app and want to import styles, those styles should not override _all_ html/body/button/input styles, but rather be scoped to the room and prejoin. 

Thought maybe the best option is actually to use a those base styles as a mixin that get's applied both in `.room-container` and `.prejoin`? Open to ideas. 

also limits `button / input` etc. to the lk-class names. I think it could be unintended to override other button / input styles just because they live within the room.


mdo (approved):
Solid improvement to scoping styles and being more explicit.

PR #315, opened 2023-02-16
Fix render loop by wrapping option in useState

Ocupe:


lukasIO:
looks good, wondering if there are other instances of arrays being passed that we should fix in the same breath?

lukasIO (approved):

packages/react/src/hooks/participant-hooks.ts
```diff
@@ -35,9 +35,8 @@ export interface UseParticipantsOptions {
  * The useParticipants hook returns all participants (local and remote) of the current room.
  */
 export const useParticipants = (options: UseParticipantsOptions = {}) => {
```
Ocupe, line 37: Removing this `useState` has no immediate effect because `useRemoteParticipants` catches the object-as-default-value problem. Still, I would argue that this is a future bug waiting to happen, and maybe the `useRemoteParticipants` hook will still get called instead of not getting called because the options object hasn't changed? 🤔

lukasIO: In the latest commit that caching is present again, had the same thought!

lukasIO (approved):

packages/react/src/components/ParticipantLoop.tsx
```diff
@@ -48,7 +48,7 @@ type ParticipantLoopProps = {
 export const ParticipantLoop = ({
   filter,
   filterDependencies,
-  updateOnlyOn,
+  updateOnlyOn = [],
```
Ocupe, line 51: This reduces the updates to the three base events. Probably not what we want.

lukasIO: hm, as long as there's no filter supplied, I would say it _is_ actually what we want

packages/react/src/hooks/participant-hooks.ts
```diff
@@ -35,13 +35,14 @@ export interface UseParticipantsOptions {
...
-  });
+  const [cachedOptions] = React.useState(options);
+  const remoteParticipants = useRemoteParticipants(cachedOptions);
   const { localParticipant } = useLocalParticipant();
 
-  return [localParticipant, ...remoteParticipants];
+  return React.useMemo(
+    () => [localParticipant, ...remoteParticipants],
+    [localParticipant, remoteParticipants],
+  );
```
Ocupe, line 45: I think wrapping this in a `useMemo` is not necessary. Testing it locally didn't have any effect for me. Different for you?

packages/react/src/components/layout/GridLayout.tsx
```diff
@@ -25,30 +25,35 @@ export interface GridLayoutProps extends React.HTMLAttributes<HTMLDivElement> {
...
-  const filterDependenciesArray = filterDependencies ?? [];
-  const filteredParticipants = React.useMemo(() => {
-    return filter ? participants.filter(filter) : participants;
-  }, [filter, participants, ...filterDependenciesArray]);
+export function GridLayout({ filter, filterDependencies = [], ...props }: GridLayoutProps) {
+  const participantOptions: UseParticipantsOptions = React.useMemo(() => {
+    return { updateOnlyOn: [] };
+  }, []);
+
+  const participants = useParticipants(participantOptions);
```
Ocupe, line 33: What is the reason for using useMemo? Could this be replaced with: 
`const participants = useParticipants({ updateOnlyOn: [] });`

lukasIO: left over from debugging, fixed

packages/react/src/hooks/participant-hooks.ts
lukasIO: left over from debugging, fixed

PR #319, opened 2023-02-16
Expose message type of usechat hook

Ocupe:


lukasIO:
hm. what I don't like about this solution is that users would have to do the `Omit<>` part themselves also, when they want to construct the messages in a type safe way.

```ts
myChatMessage: ChatMessage = {    
  ...         
}
```

lukasIO (approved):

PR #326, opened 2023-02-23
Grid-layout-revision

Ocupe:
- do not respect aspect ratio - instead optimize on filling available space
- attempt to stick to same number horizontally and vertically, because:
  - having that constraint helped to match orientation of the layout to orientation of the video. (i.e. when user is on landscape screen, 2x2 would create 4 landscape video tiles. similarly, if user is portrait, 2x2 would create 4 portrait tiles)
- sort / replacement policy is really important to get right. especially in a meetings context. we need to ensure that:
  - speaker should always be visible
  - minimize shifting of participants
    - if active speaker is already on screen, not moving them is better than putting them in the first spot
    - if they are not on screen, they should replace least-recent speaker on the first page.
- add pagination to handle participant overflow
- On rezise, don't change the row/col count independently, but switch to a smaller layout 4x4 -> 3x3 and use pageination to handle overflow.

## TODO

- [x] Automatic layout selection according to participant count and container size restrictions
- [x] Waiting on #327 to land
- [x] Finish visual stable update function


packages/core/src/sorting/sort-participants.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Default sort for participants, it'll order participants by:
+ * 1. local participant
+ * 2. dominant speaker (speaker with the loudest audio level)
+ * 3. other speakers that are recently active
+ * 4. participants with video on
+ * 5. by joinedAt
+ */
+export function sortParticipantsByVolume(participants: Participant[]): Participant[] {
```
lukasIO, line 11: is there a reason why this isn't using the base-sort-functions but rather implementing them again?

lukasIO (approved):
nice work! 

I'll bother you again about getting rid of the `GRID_LAYOUT` config, but as long as this is not something we're exposing to users, we can tackle that at a later time

packages/core/src/sorting/sort-participants.ts
Ocupe: Good catch. No, I forgot to port it back to this function.

PR #327, opened 2023-02-28
Tileloop-revision

Ocupe:


packages/react/src/hooks/track-hooks.ts
```diff
@@ -130,3 +136,93 @@ export function useTracks(sources: Array<Track.Source>, options: UseTracksOption
...
+        }
+      });
+      //TODO: Find a way to avoid type casting.
+      return pairs_ as UseTilesHookReturnType<T>;
+    } else {
+      return pairs as UseTilesHookReturnType<T>;
+    }
+  }, [pairs, participants, requirePlaceholder, sources]);
+
+  console.log('xxx', { tiles });
```
lukasIO, line 217: seems like a debug log?

packages/react/src/hooks/track-hooks.ts
```diff
@@ -130,3 +136,93 @@ export function useTracks(sources: Array<Track.Source>, options: UseTracksOption
...
+  }, [sources, participants, sources_]);
+
+  const tiles = React.useMemo<UseTilesHookReturnType<T>>(() => {
+    if (isSourcesWithOptions(sources)) {
+      const pairs_ = Array.from(pairs) as MaybeTrackParticipantPair[];
+      participants.forEach((participant) => {
+        if (requirePlaceholder.has(participant.identity)) {
+          const sourcesToAddPlaceholder = requirePlaceholder.get(participant.identity) ?? [];
+          sourcesToAddPlaceholder.forEach((placeholderSource) => {
+            console.log(
```
lukasIO, line 203: let's try to use the logger for new code that needs to log something instead of console.log

packages/core/src/observables/track.ts
```diff
@@ -75,13 +79,16 @@ export function trackParticipantPairsObservable(
   room: Room,
   sources: Track.Source[],
   options: TrackParticipantPairsObservableOptions,
-): Observable<TrackParticipantPair[]> {
+): Observable<{ trackBundles: TrackParticipantPair[]; participants: Participant[] }> {
```
lukasIO, line 82: I assume this is only temporary that you return them separately? I would have imagined all of them to be part of trackBundle now?

packages/react/src/hooks/useTracks.ts
```diff
@@ -0,0 +1,56 @@
...
+ * const pairs = useTracks(sources: [Track.Source.Camera])
+ * ```
+ */
+export function useTracks<T extends SourcesArray>(
+  sources: T,
+  options: UseTracksOptions = {},
+): UseTracksHookReturnType<T> {
+  const room = useRoomContext();
+  const [pairs, setPairs] = React.useState<TrackParticipantPair[]>([]);
+  const [, setParticipants] = React.useState<Participant[]>([]);
```
lukasIO: seems like we don't need this any more?

lukasIO (approved):

packages/core/src/observables/track.ts
Ocupe: No, this is intentional. For the placeholder thing to work, we need all participants in the room, not just those with tracks/publications. This is also the reason why all participants are returned separately.

packages/react/src/hooks/useTracks.ts
Ocupe: This was an interim status. The participants are needed to determine which participants need a placeholder.

packages/core/src/observables/track.ts
lukasIO: If the return type is a `bundle` in the format of `{participant: Participant, pub: Publication | undefined, track: Track | undefined }` it seems like we could get away with the bundle as the single return type and treat objects that do not have a publication as a placeholder?

PR #334, opened 2023-03-02
rename TrackParticipantPair to TrackBundle

Ocupe:
This PR includes two things.
## Renaming
Everything related to `TrackParticipantPair` is now called `TrackBundle`. It is no longer a pair because we have made some additions that allow us to combine published, subscribed and placeholder tracks into the same logic.

So we moved from this:
```ts
type TrackParticipantPair = {
  track: TrackPublication;
  participant: Participant;
};
```
to this:
```ts
type TrackBundleSubscribed = {
  participant: Participant;
  publication: TrackPublication;
  track: TrackPublication;
};

type TrackBundlePublished = {
  participant: Participant;
  publication: TrackPublication;
};

type TrackBundlePlaceholder = {
  participant: Participant;
  source: Track.Source;
};

export type TrackBundle = TrackBundleSubscribed | TrackBundlePublished;
export type TrackBundleWithPlaceholder =
  | TrackBundleSubscribed
  | TrackBundlePublished
  | TrackBundlePlaceholder;
```

## useTracks() revision
`useTracks()` can now return all publications, not just subscribed ones. 
```ts
const tracks = useTracks([Track.Source.Camera], {onlySubscribed: false})
```

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -16,7 +16,7 @@ export type TrackBundleSubscribed = {
   participant: Participant;
   source: Track.Source;
   publication: TrackPublication;
-  track: TrackPublication;
+  track: NonNullable<TrackPublication['track']>;
 };
 
 export type TrackBundlePublished = {
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1,7 +1,7 @@
 import type { Participant, Track, TrackPublication } from 'livekit-client';
 
 // ## PinState Type
-export type PinState = Array<TrackParticipantPair>;
+export type PinState = Array<TrackBundle>;
 export const PIN_DEFAULT_STATE: PinState = [];
 
 // ## WidgetState Types
@@ -10,8 +10,8 @@ export type WidgetState = {
 };
 export const WIDGET_DEFAULT_STATE: WidgetState = { showChat: false };
 
-// ## TrackParticipantPair Types
-export type TrackParticipantPair = {
+// ## TrackBundle Types
+export type TrackBundle = {
   track: TrackPublication;
   participant: Participant;
 };
@@ -25,12 +25,10 @@ export type TrackParticipantPlaceholder = {
   source: Track.Source;
   participant: Participant;
 };
-export type MaybeTrackParticipantPair = TrackParticipantPair | TrackParticipantPlaceholder;
+export type MaybeTrackParticipantPair = TrackBundle | TrackParticipantPlaceholder;
 
-// ### TrackParticipantPair Type Predicates
-export function isTrackParticipantPair(
-  item: MaybeTrackParticipantPair,
-): item is TrackParticipantPair {
+// ### TrackBundle Type Predicates
+export function isTrackParticipantPair(item: MaybeTrackParticipantPair): item is TrackBundle {
   return item.track !== undefined;
 }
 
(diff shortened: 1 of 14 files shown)


packages/core/src/types.ts
```diff
@@ -10,34 +10,56 @@ export type WidgetState = {
...
+export function isTrackBundleSubscribed(bundle: TrackBundle): bundle is TrackBundleSubscribed {
+  return bundle.hasOwnProperty('track');
 }
 
-export function isTrackParticipantPlaceholder(
-  item: MaybeTrackParticipantPair,
-): item is TrackParticipantPlaceholder {
-  return item.track === undefined && item.hasOwnProperty('source');
+export function isTrackBundlePublished(bundle: TrackBundle): bundle is TrackBundlePublished {
+  return bundle.hasOwnProperty('publication') && !bundle.hasOwnProperty('track');
```
lukasIO, line 51: wondering if this is the right way around.
Just from naming I would expect every `BundleSubscribed` to also be `BundlePublished`

packages/core/src/utils.ts
```diff
@@ -31,23 +36,31 @@ export const attachIfSubscribed = (
...
-    ({ track: pinnedTrack, participant: pinnedParticipant }) =>
-      pinnedTrack.trackSid === track.trackSid &&
-      pinnedParticipant.identity === participant.identity,
-  );
+  if (isTrackBundle(trackBundle)) {
+    return pinState.some(
+      (pinnedTrackBundle) =>
+        pinnedTrackBundle.participant.identity === trackBundle.participant.identity &&
+        pinnedTrackBundle.source === trackBundle.source &&
+        pinnedTrackBundle.publication.trackSid === trackBundle.publication.trackSid,
```
lukasIO, line 53: do both conditions actually have to be true here? both source and sid have to match?
If we're checking for SID I would expect that to be enough.

packages/core/src/utils.ts
```diff
@@ -31,23 +36,31 @@ export const attachIfSubscribed = (
 };
 
 /**
- * Check if the participant track is pinned.
+ * Check if the `TrackBundle` is pinned.
  */
-export function isParticipantTrackPinned(
-  trackParticipantPair: TrackParticipantPair,
+export function isTrackBundlePinned(
```
lukasIO, line 41: wondering whether this should be an overloaded method that doesn't rely on `TrackBundle` as a type.

If we leave it as `isTrackPinned` and have overloads for
```
isTrackPinned(track);
isTrackPinned(participant, source);
```
that seems slightly easier to reason about

packages/core/src/types.ts
```diff
@@ -10,34 +10,56 @@ export type WidgetState = {
...
 export const WIDGET_DEFAULT_STATE: WidgetState = { showChat: false };
 
-// ## TrackParticipantPair Types
-export type TrackParticipantPair = {
-  track: TrackPublication;
+// ## TrackBundle Types
+
+export type TrackBundleSubscribed = {
   participant: Participant;
+  source: Track.Source;
```
lukasIO, line 17: is the source needed or optional in a subscribed bundle?

packages/core/src/types.ts
Ocupe: Fair point. You would `isTrackBundle` for that. This is only to narrow down the type to the exact `TrackBundlePublished` type. But this is also not used outside so we can make it internal.

packages/core/src/utils.ts
Ocupe: Fine by me, but I would rather put it in a separate PR.

lukasIO (approved):

PR #336, opened 2023-03-02
Deprecate MediaTrack and split into AudioTrack and VideoTrack

lukasIO:


packages/react/src/components/participant/AudioTrack.tsx
```diff
@@ -0,0 +1,64 @@
...
+      volume?: number;
+    };
+
+/**
+ * The AudioTrack component is responsible for rendering participant audio tracks.
+ * This component must have access to the participant's context, or alternatively pass it a `Participant` as a property.
+ *
+ * @example
+ * ```tsx
+ * {...}
```
Ocupe, line 22: Delete `{...}`

packages/react/src/components/participant/AudioTrack.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { AudioSource, log } from '@livekit/components-core';
+import { useEnsureParticipant } from '../../context';
+import { RemoteAudioTrack } from 'livekit-client';
+
+export type AudioTrackProps<T extends HTMLMediaElement = HTMLMediaElement> =
+  React.HTMLAttributes<T> &
+    ({ source: AudioSource; name?: undefined } | { name: string; source?: undefined }) & {
+      participant?: Participant;
+      onSubscriptionStatusChanged?: (subscribed: boolean) => void;
+      volume?: number;
```
Ocupe, line 13: Maybe a small docstring that tells the user which range is required. I guess `0-1`?

packages/react/src/components/participant/MediaTrack.tsx
```diff
@@ -4,6 +4,9 @@ import { useMediaTrackBySourceOrName } from '../../hooks';
 import { ParticipantClickEvent } from '@livekit/components-core';
 import { useEnsureParticipant } from '../../context';
 
+/**
+ * @deprecated Use `AudioTrack` or `VideoTrack` instead
```
Ocupe, line 8: Perhaps a bit more drastic:
This component will be removed in the next version. Please use `AudioTrack` or `VideoTrack` instead.

Ocupe (approved):

packages/core/src/types.ts
```diff
@@ -82,3 +82,6 @@ export type RequireOnlyOne<T, Keys extends keyof T = keyof T> = Pick<T, Exclude<
   {
     [K in Keys]-?: Required<Pick<T, K>> & Partial<Record<Exclude<Keys, K>, undefined>>;
   }[Keys];
+
+export type AudioSource = Track.Source.Microphone | Track.Source.ScreenShareAudio;
+export type VideoSource = Track.Source.Camera | Track.Source.ScreenShare;
```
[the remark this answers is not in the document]
lukasIO: I had included it at first, but removed, because I thought it might be better to make it explicit for those types. For `Unknown` we do not know by definition whether it is an AudioSource or a VideoSource.
The advantage of leaving it out, is that we avoid overlap

packages/react/src/components/participant/AudioTrack.tsx
lukasIO: generally it's 0-1, however users have the possibility to use the webAudio option in which case the range can also be user defined. Will add a docstring that says `by default the range is between 0 and 1`

PR #343, opened 2023-03-08
move GridLayout logic into a hooks

Ocupe:


packages/react/src/hooks/useGridLayout.ts
```diff
@@ -0,0 +1,45 @@
...
+ * If the available screen space is not enough, it will reduce the number of maximum visible
+ * tiles and select a layout that still works visually within the given limitations.
+ * As the order of tiles changes over time, the hook tries to keep visual updates to a minimum
+ * while trying to display important tiles such as speaking participants or screen shares.
+ */
+export function useGridLayout(
+  /** HTML element that contains the gird. */
+  containerElement: React.RefObject<HTMLDivElement>,
+  /** Direct child of the `containerElement`. */
+  gridElement: React.RefObject<HTMLDivElement>,
```
lukasIO, line 22: could we automatically detect that `gridElement` instead of having to pass both explicitly?

packages/react/src/hooks/useVisualStableUpdate.ts
```diff
@@ -0,0 +1,55 @@
...
+ * they will be moved to the first page by replacing the least active/interesting participant on the first page.
+ */
+export function useVisualStableUpdate(
+  /** `TrackBundle`s to display in the grid.  */
+  trackBundles: TrackBundleWithPlaceholder[],
+  maxItemsOnPage: number,
+  options: UseVisualStableUpdateOptions = {},
+): TrackBundleWithPlaceholder[] {
+  const stateTrackBundles = React.useRef<TrackBundleWithPlaceholder[]>([]);
+  const maxTilesOnPage = React.useRef<number>(-1);
```
lukasIO, line 28: what's the reason for adapting `useRef`s here vs. `useState`?

packages/react/src/hooks/useGridLayout.ts
Ocupe: Good idea. I wonder if we can get by with just one element? Or would it be a problem if the grid is also applied to the containing element?

lukasIO: not sure, but worth a try!

packages/react/src/hooks/useVisualStableUpdate.ts
lukasIO: makes sense! thanks :)

PR #344, opened 2023-03-08
add usePagination hook

Ocupe:
This hook is to be used with both the GridLayout and the Carousel component

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -58,14 +58,18 @@
     "@size-limit/file": "^8.1.2",
     "@size-limit/webpack": "^8.1.2",
     "@svgr/cli": "^6.5.1",
+    "@testing-library/react": "^14.0.0",
     "@types/react": "^18.0.25",
     "@types/react-dom": "^18.0.8",
+    "@vitejs/plugin-react": "^3.1.0",
     "eslint-config-lk-custom": "~0.1.1",
+    "jsdom": "^21.1.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "size-limit": "^8.1.2",
     "tsup": "^6.5.0",
-    "typescript": "^4.9.4"
+    "typescript": "^4.9.4",
+    "vite": "^4.1.4"
   },
   "engines": {
     "node": ">=14"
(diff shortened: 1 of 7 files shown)


packages/react/package.json
```diff
@@ -58,14 +58,18 @@
...
+    "@vitejs/plugin-react": "^3.1.0",
     "eslint-config-lk-custom": "~0.1.1",
+    "jsdom": "^21.1.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "size-limit": "^8.1.2",
     "tsup": "^6.5.0",
-    "typescript": "^4.9.4"
+    "typescript": "^4.9.4",
+    "vite": "^4.1.4"
```
lukasIO, line 72: is `vitest` is missing as a dependency here?

packages/react/src/hooks/usePagination.ts
```diff
@@ -0,0 +1,55 @@
...
+      } else {
+        if (state === 1) {
+          return state;
+        }
+        return state - 1;
+      }
+    });
+  };
+
+  const setPage_ = (num: number) => {
```
lukasIO, line 34: very personal opinion: naming feels a bit awkward with the underscore. `goToPage` maybe?

packages/react/src/hooks/usePagination.ts
```diff
@@ -0,0 +1,55 @@
...
+ * The `usePagination` hook implements simple pagination logic for use with arrays.
+ * @example
+ * ```tsx
+ * const { firstItemIndex, lastItemIndex } = usePagination(3, 9);
+ * const itemsOnPage = items.slice(firsItemIndex, lastItemIndex)
+ * ```
+ * @internal
+ */
+export function usePagination(itemPerPage: number, totalItemCount: number) {
+  const [page, setPage] = useState(1);
```
lukasIO, line 13: wondering if we should rename this to `currentPage` and return all other pages as well? 🤔

Ocupe: `currentPage` is fine by my 👍 
What do you mean by "return all other pages"? At the moment only the index of the page is returned.

lukasIO: ah, you're right. I assumed it would return the relevant trackBundles directly. But index is fine by me!

PR #346, opened 2023-03-09
CraouselView revision

Ocupe:
Update the CraouselView component:
- Add a visually stable update to the carousel component.
- Limit the maximum visible tiles by container size, height, and minimum tile width or height.
- Add snap scrolling to better align the tiles.

### DIFF
--- a/packages/react/src/components/layout/CarouselView.tsx
+++ b/packages/react/src/components/layout/CarouselView.tsx
@@ -28,19 +28,22 @@ export function CarouselView({ filter, filterDependencies = [], ...props }: Caro
   }, [filter, layoutContext?.pin.state, trackBundles, ...filterDependencies]);
 
   const { width, height } = useSize(asideEl);
-  const isVertical = height >= width;
-  const maxVisibleTiles = isVertical
-    ? Math.max(1, Math.floor(height / MIN_HEIGHT))
-    : Math.floor(width / MIN_WIDTH);
+  const orientation = height >= width ? 'vertical' : 'horizontal';
+  const maxVisibleTiles =
+    orientation === 'vertical'
+      ? Math.max(1, Math.floor(height / MIN_HEIGHT))
+      : Math.floor(width / MIN_WIDTH);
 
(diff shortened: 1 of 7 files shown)


lukasIO (approved):

packages/react/src/components/layout/CarouselView.tsx
```diff
@@ -0,0 +1,54 @@
...
+  ]);
+  const filteredTiles = React.useMemo(() => {
+    const tilesWithoutPinned = trackBundles.filter(
+      (tile) => !layoutContext?.pin.state || !isTrackBundlePinned(tile, layoutContext.pin.state),
+    );
+    return filter ? tilesWithoutPinned.filter(filter) : tilesWithoutPinned;
+  }, [filter, layoutContext?.pin.state, trackBundles, ...filterDependencies]);
+
+  const { width, height } = useSize(asideEl);
+  const isVertical = height >= width;
```
lukasIO, line 31: I think this is good as a guess, but maybe better if we allow to (additionally) override this via a prop setting?

packages/styles/scss/components/layout/_carousel-view.scss
```diff
@@ -0,0 +1,45 @@
+.carousel {
+  --carousel-is-vertical: 1;
+  --carousel-is-horizontal: 0;
```
lukasIO, line 3: these are mutually exclusive. would be nicer if we could represent them as a single var. 
thinking whether it might be more in line with our other settings if this was in fact a data attribute?

lukasIO (approved):

packages/styles/scss/components/layout/_carousel-view.scss
Ocupe: Because it is CSS, inverting is not a simple operation. But now I derive the horizontal value from the vertical and only have to set `--carousel-is-vertical`

packages/react/src/components/layout/CarouselView.tsx
Ocupe: Good point, but since we would also have to change some things in the css, I'll create a new issue for that.

PR #347, opened 2023-03-09
Fix participant tile showing up twice

lukasIO:


packages/react/src/hooks/useTracks.ts
```diff
@@ -68,6 +69,14 @@ export function useTracks<T extends SourcesArray>(
         if (requirePlaceholder.has(participant.identity)) {
           const sourcesToAddPlaceholder = requirePlaceholder.get(participant.identity) ?? [];
           sourcesToAddPlaceholder.forEach((placeholderSource) => {
+            if (
+              trackBundles.find(
+                ({ participant: p, publication }) =>
+                  participant.identity === p.identity && publication.source === placeholderSource,
+              )
+            ) {
+              return;
```
lukasIO, line 78: this is the main change here, if there's already a publication present, don't add the placeholder

Ocupe (approved):

PR #352, opened 2023-03-14
UPDATE: Make useDataChannel hook except generic string types instead …

brkcrdk:
Make useDataChannel hook except generic string types instead of just string. This way we can add type safety to our message topics. We can use `useDataChannel` like this as well:
```js
type ChannelMessageType = 'SAY_HI' | 'SAY_BYE';

// works
const { isSending } = useDataChannel<ChannelMessageType>('SAY_HI'); 
// TS warns us 
const { isSending } = useDataChannel<ChannelMessageType>('SAYY_HI');
```

### DIFF
--- a/packages/react/src/hooks/dataChannel-hooks.ts
+++ b/packages/react/src/hooks/dataChannel-hooks.ts
@@ -3,7 +3,7 @@ import * as React from 'react';
 import { useRoomContext } from '../context';
 import { useObservableState } from '../helper';
 
-export function useDataChannel(topic?: string) {
+export function useDataChannel<T extends string>(topic?: T) {
   const room = useRoomContext();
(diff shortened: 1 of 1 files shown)


lukasIO:
hi @brkcrdk, 
thanks for the PR! 

The reason we removed this was that we were no longer able to meaningfully parse the incoming messages while keeping compatibility to other SDKs.

The pattern we had in mind for making sure that you retain some sort of type safety when it comes to passing in a topic name would be
```ts
const MyTopics = {
   SAY_HI: 'sayhi',
   SAY_BYE: 'saybye'
} as const;

useDataChannel(MyTopics.SAY_HI);
```

brkcrdk:
Oh, I see. That's makes sense. :D 

Thank you for sharing. I will use the way you suggested then. 🙏🏻

PR #374, opened 2023-03-20
Enable trackReferences to be used on audio/video track

lukasIO:
allows for usage of `Audio/VideoTrack` like this:

```tsx

export const AllMicRenderer = () => {

  const trackReferences = useTracks([Track.Source.Microphone]);

  return (
    <div style={{ display: 'none' }}>
      {trackReferences.map((trackRef) => (
        <AudioTrack key={trackRef.publication.trackSid} trackReference={trackRef} />
      ))}
    </div>
  );
};

```

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -38,12 +38,17 @@ export interface ParticipantClickEvent {
   track?: TrackPublication;
 }
 
-export type TrackObserverOptions =
-  | { source: Track.Source; name?: undefined }
-  | { source?: undefined; name: string };
+export type TrackSource<T extends Track.Source> = RequireAtLeastOne<
+  { source: T; name: string },
+  'name' | 'source'
+>;
+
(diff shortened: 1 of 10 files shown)


packages/core/src/types.ts
```diff
@@ -38,12 +38,17 @@ export interface ParticipantClickEvent {
...
 
-export type TrackObserverOptions =
-  | { source: Track.Source; name?: undefined }
-  | { source?: undefined; name: string };
+export type TrackSource<T extends Track.Source> = RequireAtLeastOne<
+  { source: T; name: string },
+  'name' | 'source'
+>;
+
+export type TrackIdentifier<T extends Track.Source = Track.Source> =
```
davidzhao, line 46: nit: would be helpful to include a bit of comments showing examples of what would qualify. Since I imagine users would click into `TrackIdentifier` and try to understand what it means.

davidzhao (approved):
lgtm!

PR #388, opened 2023-03-29
fix: spread track to ParticipantTile

Ocupe:
fix error when going into focus view


lukasIO:
superseded by https://github.com/livekit/components-js/pull/389

PR #395, opened 2023-03-31
Fix chat toggle not working

lukasIO:


packages/react/src/components/layout/LayoutContextProvider.tsx
```diff
@@ -1,55 +1,32 @@
...
   React.useEffect(() => {
     log.debug('PinState Updated', { state: layoutContextValue.pin.state });
     if (onPinChange && layoutContextValue.pin.state) onPinChange(layoutContextValue.pin.state);
   }, [layoutContextValue.pin.state, onPinChange]);
 
   React.useEffect(() => {
-    log.debug('Widget Updated', { widgetState });
-    if (onWidgetChange) onWidgetChange(widgetState);
-  }, [onWidgetChange, widgetState]);
+    log.debug('Widget Updated', { widgetState: layoutContextValue.widget.state });
```
lukasIO, line 25: this was the main problem, the widget effect was still relying on an unused value of the layout context. 
refactored around it so that the mistake cannot be made any more

packages/react/src/context/layout-context.ts
```diff
@@ -38,6 +39,26 @@ export function useEnsureLayoutContext(layoutContext?: LayoutContextType) {
...
+export function useCreateLayoutContext(): LayoutContextType {
+  const [pinState, pinDispatch] = React.useReducer(pinReducer, PIN_DEFAULT_STATE);
+  const [widgetState, widgetDispatch] = React.useReducer(chatReducer, WIDGET_DEFAULT_STATE);
+  return {
+    pin: { dispatch: pinDispatch, state: pinState },
+    widget: { dispatch: widgetDispatch, state: widgetState },
+  };
+}
+
+export function useEnsureCreateLayoutContext(layoutContext?: LayoutContextType): LayoutContextType {
```
lukasIO, line 51: open for better naming suggestions 👀

Ocupe (approved):
Thanks for the fix 🙏

PR #397, opened 2023-03-31
Chat-revision

Ocupe:
- Group consecutive chat messages from the same participant. Show only the name on the first message.
- Increase default chat width to work better in focus layout.
- update chat message styling
- add prop to ChatEntry to allow hiding chat message meta data (name, timestamp).


lukasIO (approved):
🔥

mdo:
Oh hell yeah, this is awesome!

PR #414, opened 2023-04-03
Apply resets only on lk- class elements

### LINKED ISSUE #384: Conflicting styles from @livekit/components-styles and @mui

\@livekit/components-styles

### Describe the bug

When importing styles from @livekit/components-styles, styles shouldn't be polluting the global scope.

Because of this, styles imported conflict with other packages like @mui

This is the example with the style conflicting with the textfield
Screenshot 2023-03-28 at 22 00 24

This is the example with the livekit style disabled:
Screenshot 2023-03-28 at 22 00 37

### Reproduction

Use @mui and @livekit in a project -> implement textfield

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 13.2.1
    CPU: (8) arm64 Apple M2
    Memory: 200.45 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.0 - ~/.nvm/versions/node/v16.17.0/bin/node
    Yarn: 1.22.19 - ~/Documents/xxxx/node_modules/.bin/yarn
    npm: 8.19.1 - ~/.nvm/versions/node/v16.17.0/bin/npm
  Browsers:
    Chrome: 111.0.5563.146
    Safari: 16.3
  npmPackages:
    livekit-client: ^1.7.1 => 1.7.1
```

### Severity

blocking an upgrade

### Additional Information

_No response_

lukasIO:
fixes #384

### DIFF
--- a/packages/styles/scss/components/_room.scss
+++ b/packages/styles/scss/components/_room.scss
@@ -6,7 +6,8 @@
   width: 100%;
   height: 100%;
 
-  * {
+  *[class^='lk-'],
+  *[class*=' lk-'] {
     box-sizing: border-box;
   }
 }
--- a/packages/styles/scss/themes/default.scss
+++ b/packages/styles/scss/themes/default.scss
@@ -60,5 +60,5 @@ $accent: #1f8cf9;
   --box-shadow: 0 0.5rem 1.5rem rgba(0, 0, 0, 0.15);
 
   --grid-gap: 0.5rem;
-  --control-bar-height: 62px;
+  --control-bar-height: 69px;
 }
--- a/.changeset/neat-moons-relax.md
+++ b/.changeset/neat-moons-relax.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-styles": patch
+---
+
+Apply resets only on lk- class elements
--- a/packages/styles/postcss.config.js
+++ b/packages/styles/postcss.config.js
@@ -4,7 +4,7 @@ module.exports = {
   plugins: [
     require('postcss-prefixer')({
       prefix: `${PREFIX}-`,
-      ignore: [/\[class\*=" lk-"\]/],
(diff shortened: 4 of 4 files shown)


packages/styles/scss/themes/default.scss
```diff
@@ -60,5 +60,5 @@ $accent: #1f8cf9;
   --box-shadow: 0 0.5rem 1.5rem rgba(0, 0, 0, 0.15);
 
   --grid-gap: 0.5rem;
-  --control-bar-height: 62px;
+  --control-bar-height: 69px;
```
mdo, line 63: Made this adjustment for you to fix a little bit of white space that was showing through on video conference view. The control bar's height has increased, perhaps as a result of some padding box changes? Still reviewing.

packages/styles/scss/components/_room.scss
```diff
@@ -6,7 +6,8 @@
   width: 100%;
   height: 100%;
 
-  * {
+  *[class^='lk-'],
+  *[class*=' lk-'] {
```
mdo, line 10: If we're using the `attr*=''` selector here, I think we could get rid of the prefixed attribute selector and simplify this a bit. If an element has a class containing `lk-`, that should always match our conditions I think?

```suggestion
  *[class*='lk-'] {
```

lukasIO: having both in there matches for
```html
<!-- match -->
<div class="lk-something" />
<!-- match -->
<div class="my-class lk-something" />
<!-- no match -->
<div class="bulk-something" />
```

the last example with "bulk-something" would still match if we did _only_  `*[class*='lk-']`

lukasIO: but yeah, only using this pattern as a quick fix. If there's a cleaner solution you have in mind, feel free to change it up and kick this out entirely!

Ocupe (approved):

PR #421, opened 2023-04-05
listen also to ActiveSpeakersChanged event in VideoConference

Ocupe:
This PR fixes that the VideoConference component is not updated when a participant starts speaking.


davidzhao:
if someone was using `useVisualStableUpdate` with a list of tracks, would what's returned automatically update when there are speaker changes?

lukasIO:
yup, this was only a problem because we passed in the empty array in `updateOnlyOn`

PR #422, opened 2023-04-06
fix carousel comp

Ocupe:
- undo a css hack to prevent unstable UI states in CarouselView. Unfortunately it did not work on iOS
- handle the unstable UI state by taking into account the width of the scrollbar when CarouselView overflows.

- [x] CarouselView does not update to show talking participant

### DIFF
--- a/packages/react/src/components/layout/CarouselView.tsx
+++ b/packages/react/src/components/layout/CarouselView.tsx
@@ -36,11 +36,13 @@ export function CarouselView({ tracks, orientation, ...props }: CarouselViewProp
       : Math.max(Math.floor(width / tileWidth), MIN_VISIBLE_TILES);
 
   // To avoid an unstable UI state, on overflow, we need to consider the scrollbar wide.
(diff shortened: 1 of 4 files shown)


packages/react/src/components/layout/CarouselView.tsx
```diff
@@ -1,4 +1,4 @@
-import type { TrackReferenceOrPlaceholder } from '@livekit/components-core';
+import { TrackReferenceOrPlaceholder, getScrollBarWidth } from '@livekit/components-core';
```
lukasIO, line 1: I think `TrackReferencePlaceholder` should still be a type import

lukasIO (approved):

packages/react/src/components/layout/CarouselView.tsx
Ocupe: Well spotted! Prettiere and Autoimport can mess up the imports.

PR #423, opened 2023-04-06
add swipe to grid

Ocupe:


packages/react/src/hooks/useSwipe.ts
```diff
@@ -0,0 +1,54 @@
...
+ * ```tsx
+ *  <div
+      onTouchStart={onTouchStart}
+      onTouchMove={onTouchMove}
+      onTouchEnd={onTouchEnd}
+    >
+ * @internal
+ * ```
+ */
+export function useSwipe(
```
lukasIO, line 20: just as a suggestion: 
```
useSwipe(elementReference, onSwipe)
```

feels a lot simpler api wise and all the logic of registering events could be handled inside the hook. 
`onSwipe` would include a `direction` parameter, so we don't have to change the API if we need to add more swipe directions later on

PR #428, opened 2023-04-11
Examples revision

Ocupe:
- update examples when needed
- update examples overview page (add logo)
- remove broken examples


examples/nextjs/pages/clubhouse.tsx
```diff
@@ -31,7 +29,7 @@ const Clubhouse = () => {
   const [connected, setConnected] = useState(false);
 
   return (
-    <div className={styles.container}>
+    <div data-lk-theme="default" className={styles.container}>
```
lukasIO, line 32: looks like the example doesn't render any audio?

PR #429, opened 2023-04-11
fix style participant audio tile

Ocupe:
The black background of the connection indicator was always visible.

before: 
Screenshot 2023-04-11 at 12 13 25

after:
Screenshot 2023-04-11 at 12 13 40


lukasIO (approved):
only downside is that if someone chose to change the bg-colours there wouldn't be a way to separate contrasts between connection indicator and background, right?

Ocupe:
> only downside is that if someone chose to change the bg-colours there wouldn't be a way to separate contrasts between connection indicator and background, right?

I'm not sure I follow. You can still set the background color of the connection quality icon if that answers your question.

lukasIO:
yeah, that's what I meant, nice!

PR #431, opened 2023-04-12
add pagination indicator to gird

Ocupe:


packages/react/src/components/controls/PaginationIndicator.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+
+export type PaginationIndicatorProps = {
+  totalPageCount: number;
+  currentPage: number;
+};
+
+export function PaginationIndicator({ totalPageCount, currentPage }: PaginationIndicatorProps) {
+  console.log('xxx', { totalPageCount }, { currentPage });
```
lukasIO, line 9: remove?

PR #434, opened 2023-04-13
inside tile speaking indicator

Ocupe:
This PR moves the speaking indicator inside the tile to work with GridLayout and Carousel (desktop and mobile). 
Fixed:
- speaking indicator cut off
- small gap in corner between speaking indicator and video element

### DIFF
--- a/packages/styles/scss/components/participant/_participant-tile.scss
+++ b/packages/styles/scss/components/participant/_participant-tile.scss
@@ -6,17 +6,26 @@
   gap: 0.375rem;
   overflow: hidden;
   border-radius: var(--border-radius);
-  transition-property: box-shadow;
-  transition-timing-function: ease-in-out;
-  // When participant stops talking the transition longer and starts with a delay.
-  transition-delay: 0.5s;
-  transition-duration: 0.8s;
-  box-shadow: 0px 0px 0px 0 transparent;
-  &[data-speaking='true']:not([data-source='screen_share']) {
-    // When participant starts talking the transition short and without delay.
+  &::after {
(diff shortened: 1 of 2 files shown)


packages/styles/scss/components/participant/_participant-tile.scss
```diff
@@ -6,17 +6,26 @@
...
   border-radius: var(--border-radius);
-  transition-property: box-shadow;
-  transition-timing-function: ease-in-out;
-  // When participant stops talking the transition longer and starts with a delay.
-  transition-delay: 0.5s;
-  transition-duration: 0.8s;
-  box-shadow: 0px 0px 0px 0 transparent;
-  &[data-speaking='true']:not([data-source='screen_share']) {
-    // When participant starts talking the transition short and without delay.
+  &::after {
```
lukasIO, line 9: out of curiosity: why in `:after` vs. just on the class itself?

Ocupe: `::after` overlays the tile and therefore does not change the size of the tile content. If we put it on the class directly it would scale the video when the indicator shows up.
Another way would be to go with a invisible border and then change the color or opacity on speaking. This can introduce alignment issues. For example, when you want the tile be flush against some "design"-line but the invisible border prevents that.

If we ever want to support switching between the outside and inside indicators. This approach would give us the flexibility to simply add twice the border width to the width and height of the :`:after` overlay to achieve an outside speaking indicator. To switch between the two, we could use a data-attr or property.

lukasIO (approved):

PR #435, opened 2023-04-13
Remove filter from ParticipantLoop 

Ocupe:
This PR changes:
- update participant hooks to accept a room
- remove filter from ParticipantLoop
- updates examples that use the ParticipantLoop comp


lukasIO (approved):
lgtm! a changeset would be nice!

PR #446, opened 2023-04-18
Track ref all the things

Ocupe:
- Use `useEnsureTrackReference` instead of `useEnsureParticipant`


packages/core/src/track-reference/track-reference.utils.ts
```diff
@@ -62,3 +62,29 @@ export function isTrackReferencePinned(
...
 }
+
+/**
+ * Simple way to create a `TrackReference` from its parts.
+ */
+export function trackReference(
+  participant: Participant,
+  source: Track.Source,
+  publication?: TrackPublication,
+  track?: Track,
```
lukasIO, line 73: is this needed for types? otherwise we could always infer `track` from the publication

PR #449, opened 2023-04-18
Replace lodash array functions with custom functions

lukasIO:


packages/core/src/sorting/tile-array-update.ts
```diff
@@ -9,13 +8,27 @@ type VisualChanges<T> = {
...
+}
+
+function zip<T, U>(a1: Array<T>, a2: Array<U>) {
+  return a1.map((value, idx) => [value, a2[idx]]);
+}
+
+function differenceBy<T>(a1: Array<T>, a2: Array<T>, by: (arg: T) => string) {
+  return a1.filter((c) => !a2.map((v) => by(v)).includes(by(c)));
+}
+
```
Ocupe, line 24: This seems to general to be placed here. Maybe move it into a internal `helper/array-extra.ts` file?

Ocupe (approved):

PR #451, opened 2023-04-19
Shrink bundle size for LiveKitRoom

lukasIO:
If users decide to _only_ import `LiveKitRoom` we want to try to use as little resources as possible. 
Some of our dependencies were either unnecessary or caused the bundle to bloat even when not being imported (failed tree shaking). This PR addresses the lowest hanging fruits in that regard so that importing `LiveKitRoom` by itself only takes ~1.9kB.


Ocupe (approved):
👍

PR #455, opened 2023-04-19
Default username displayed in input box

apurvbhavsar:
### Improvement

Manually added username in PreJoin Component using props not shown in Input Box. 
Fixed that issue.

### DIFF
--- a/.changeset/six-rice-perform.md
+++ b/.changeset/six-rice-perform.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-react": patch
+---
+
+Display default username in PreJoin component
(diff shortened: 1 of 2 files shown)


lukasIO:
Hi @apurvbhavsar ,
thanks for the PR! 
This makes a lot of sense in general, just thinking that we should probably go with `defaultValue` instead of `value` (https://legacy.reactjs.org/docs/uncontrolled-components.html#default-values).

apurvbhavsar:
Yes, it will be nice and solve the issue too..

Thanks, Let me update it again

lukasIO (approved):
great! thank you!

Ocupe (approved):
Good eye for detail, thank you. 🎉

PR #463, opened 2023-04-20
set log level to error

Ocupe:


packages/core/src/logger.ts
```diff
@@ -1,6 +1,6 @@
 import loglevel from 'loglevel';
 
 const log = loglevel.getLogger('lk-components-js');
-log.setDefaultLevel('DEBUG');
+log.setDefaultLevel('ERROR');
```
lukasIO, line 4: how about `warn` ?

Ocupe: Now I have to change the branch name and the PR title :D
Warning is fine by me 👍

lukasIO (approved):

PR #465, opened 2023-04-21
Cleanup deprecated stuff

Ocupe:
This PR removes deprecated components:

- MediaTrack
- ScreenShareRenderer

MediaTrack is replaced by AudioTrack and VideoTrack in a way that mimics the behavior of MediaTrack before. There is still some ambiguity about how to handle tracks with an unknown source.


packages/react/src/components/participant/ParticipantTile.tsx
```diff
@@ -140,13 +141,21 @@ export const ParticipantTile = ({
...
         {children ?? (
           <>
-            {/** TODO remove MediaTrack in favor of the equivalent Audio/Video Track. need to figure out how to differentiate here */}
-            <MediaTrack
-              source={source}
-              publication={publication}
-              participant={participant}
-              onSubscriptionStatusChanged={handleSubscribe}
-            />
+            {source === Track.Source.Camera || source === Track.Source.ScreenShare ? (
```
lukasIO, line 144: unfortunately this won't handle all cases. 
If we have a publication present we should be using `publication.kind` to determine the right kind of track. 

This also poses the question whether we want to do `track: TrackReferenceOrPlaceholder` instead of spreading the arguments out. This has some ugly side-effects (like implicit context usage will get harder for users), so I'm not totally convinced

Ocupe: I agree, I think it still treats it like the old logic. If it was not a `camera` or `screen_share` source, it would always treat it like an audio track.

lukasIO: I think adding `publication?.kind === 'video'` would be a rather low hanging fruit?

lukasIO (approved):

PR #471, opened 2023-04-27
Prevent false warning when setting volume on AudioTrack

Ocupe:
This PR should fix the bug report from our community: https://livekit-users.slack.com/archives/C01KZ61HZRR/p1682606143554289

Thanks to Andres Garcia for bringing this up 🙏

### DIFF
--- a/packages/react/src/hooks/useMediaTrackBySourceOrName.ts
+++ b/packages/react/src/hooks/useMediaTrackBySourceOrName.ts
@@ -16,7 +16,7 @@ export function useMediaTrackBySourceOrName(
   const [publication, setPublication] = React.useState(getTrackByIdentifier(observerOptions));
 
   const [isMuted, setMuted] = React.useState(publication?.isMuted);
-  const [isSubscribed, setSubscribed] = React.useState(publication?.isSubscribed);
+  const [isSubscribed, setSubscribed] = React.useState(publication?.isSubscribed ?? false);
   const [track, setTrack] = React.useState(publication?.track);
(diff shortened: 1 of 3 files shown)


packages/react/src/hooks/useMediaTrackBySourceOrName.ts
```diff
@@ -16,7 +16,7 @@ export function useMediaTrackBySourceOrName(
   const [publication, setPublication] = React.useState(getTrackByIdentifier(observerOptions));
 
   const [isMuted, setMuted] = React.useState(publication?.isMuted);
-  const [isSubscribed, setSubscribed] = React.useState(publication?.isSubscribed);
+  const [isSubscribed, setSubscribed] = React.useState(publication?.isSubscribed ?? false);
```
lukasIO, line 19: I guess, this changes the type of `isSubscribed` from boolean to `boolean | undefined`. 
can we avoid/revert that?

packages/react/src/hooks/useMediaTrackBySourceOrName.ts
```diff
@@ -16,7 +16,7 @@ export function useMediaTrackBySourceOrName(
   const [publication, setPublication] = React.useState(getTrackByIdentifier(observerOptions));
 
   const [isMuted, setMuted] = React.useState(publication?.isMuted);
```
lukasIO, line 18: should we do the same thing for isMuted ?

lukasIO (approved):

PR #487, opened 2023-05-22
add useAudioPlayback hook and expose existing hooks

Ocupe:
Add `usePreviewDevice`, `useStartAudio` and newly created `useAudioPlayback` hooks to public API.

### DIFF
--- a/packages/react/etc/components-react.api.md
+++ b/packages/react/etc/components-react.api.md
@@ -427,6 +427,12 @@ export type TrackToggleProps<T extends ToggleSource> = Omit<React_2.ButtonHTMLAt
     captureOptions?: CaptureOptionsBySource<T>;
 };
 
+// @alpha
+export function useAudioPlayback(room: Room): {
+    canPlayAudio: boolean;
+    startAudio: () => Promise<void>;
+};
+
 // @public (undocumented)
 export function useChat(): {
     send: ((message: string) => Promise<void>) | undefined;
@@ -705,10 +711,18 @@ export function useSortedParticipants(participants: Array<Participant>): Partici
 export const useSpeakingParticipants: () => Participant[];
 
 // @alpha
-export function useStartAudio(room: Room): {
+export function useStartAudio({ room, props }: UseStartAudioProps): {
+    mergedProps: React_2.HTMLAttributes<HTMLElement>;
     canPlayAudio: boolean;
-    startAudio: () => void;
 };
+
+// @alpha (undocumented)
+export interface UseStartAudioProps {
+    // (undocumented)
+    props: React_2.ButtonHTMLAttributes<HTMLButtonElement>;
+    // (undocumented)
+    room: Room;
+}
 
 // @alpha
 export function useSwipe(element: React_2.RefObject<HTMLElement>, options?: UseSwipeOptions): void;
--- a/packages/react/etc/components-react.api.md
+++ b/packages/react/etc/components-react.api.md
@@ -428,7 +428,7 @@ export type TrackToggleProps<T extends ToggleSource> = Omit<React_2.ButtonHTMLAt
 };
 
 // @alpha
-export function useAudioPlayback(room: Room): {
+export function useAudioPlayback(room?: Room): {
     canPlayAudio: boolean;
     startAudio: () => Promise<void>;
 };
@@ -721,7 +721,7 @@ export interface UseStartAudioProps {
     // (undocumented)
     props: React_2.ButtonHTMLAttributes<HTMLButtonElement>;
     // (undocumented)
-    room: Room;
+    room?: Room;
 }
 
 // @alpha
--- a/packages/react/etc/components-react.api.md
+++ b/packages/react/etc/components-react.api.md
@@ -15,8 +15,10 @@ import { ConnectionState as ConnectionState_2 } from 'livekit-client';
 import { DataSendOptions } from '@livekit/components-core';
 import type { GridLayout as GridLayout_2 } from '@livekit/components-core/dist/helper/grid-layouts';
 import { HTMLAttributes } from 'react';
+import type { LocalAudioTrack } from 'livekit-client';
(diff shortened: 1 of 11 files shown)


packages/react/src/components/controls/StartAudio.tsx
```diff
@@ -5,35 +5,32 @@ import { useRoomContext } from '../../context';
...
 
-interface UseStartAudioProps {
-  room: Room;
-  props: React.ButtonHTMLAttributes<HTMLButtonElement>;
-}
-
-export function useStartAudio({ room, props }: UseStartAudioProps) {
-  const { className, roomAudioPlaybackAllowedObservable, handleStartAudioPlayback } = React.useMemo(
+/**
+ * In many browser to start audio playback, the user must perform a user-initiated event such as clicking a button.
```
lukasIO, line 9: nit: missing `s` on browsers.

packages/react/src/components/controls/StartAudio.tsx
```diff
@@ -5,35 +5,32 @@ import { useRoomContext } from '../../context';
...
+ * The `useStatAudio` hook returns an object with a boolean `canPlayAudio` flag
```
lukasIO, line 10: nit: typo

packages/react/src/components/controls/StartAudio.tsx
```diff
@@ -5,35 +5,32 @@ import { useRoomContext } from '../../context';
...
+ * that indicates whether audio playback is allowed in the current context,
+ * as well as a `startAudio` function that can be called in a button `onClick` callback to start audio playback in the current context.
+ *
+ * @see Autoplay policy on MDN web docs for more info: {@link https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Best_practices#autoplay_policy}
+ * @alpha
+ */
+export function useStartAudio(room: Room) {
+  const { roomAudioPlaybackAllowedObservable, handleStartAudioPlayback } = React.useMemo(
```
lukasIO, line 18: it's a bit unfortunate that both (this hook and the `StartAudio` button) call the same function `setupStartAudio`. Wondering if we can avoid that.

packages/react/src/components/controls/StartAudio.tsx
```diff
@@ -5,35 +5,32 @@ import { useRoomContext } from '../../context';
...
+/**
+ * In many browser to start audio playback, the user must perform a user-initiated event such as clicking a button.
+ * The `useStatAudio` hook returns an object with a boolean `canPlayAudio` flag
+ * that indicates whether audio playback is allowed in the current context,
+ * as well as a `startAudio` function that can be called in a button `onClick` callback to start audio playback in the current context.
+ *
+ * @see Autoplay policy on MDN web docs for more info: {@link https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Best_practices#autoplay_policy}
+ * @alpha
+ */
+export function useStartAudio(room: Room) {
```
lukasIO, line 17: I also like `useAudioPlayback` as a more general term. the returned `canPlayAudio` makes more sense with that name imo

lukasIO (commented):
just a couple of hopefully minor things

packages/react/src/components/controls/StartAudio.tsx
Ocupe: `useAudioPlayback` is a better name, but we would break the naming convention of the component and the hook having the same name. I would then move the `useAudioPlayback` component to the hooks folder and remove it from here. What do you think about that?

packages/react/src/components/controls/StartAudio.tsx
Ocupe: The alternative was to hardcode the class name. We could have a gernall hook without the class name and a useStartAudio that wraps the gernall hook, merges props and adds the class name.
Or do you have another suggestion?

lukasIO: I like the approach of splitting it in two! but that would also mean changing core API right?

packages/react/src/components/controls/StartAudio.tsx
lukasIO: that sounds good! (but also seems like if we solve the issue above with two hooks, the naming problem would fix itself because we'd have a general hook anyways)

packages/react/src/components/controls/StartAudio.tsx
Ocupe: As it turned out, no change to the core API was necessary.

packages/react/src/hooks/useAudioPlayback.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * In many browsers to start audio playback, the user must perform a user-initiated event such as clicking a button.
+ * The `useAudioPlayback` hook returns an object with a boolean `canPlayAudio` flag that indicates whether audio
+ * playback is allowed in the current context, as well as a `startAudio` function that can be called in a button
+ * `onClick` callback to start audio playback in the current context.
+ *
+ * @see Autoplay policy on MDN web docs for more info: {@link https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Best_practices#autoplay_policy}
+ * @alpha
+ */
+export function useAudioPlayback(room: Room): {
```
lukasIO, line 15: should room be optional and `useEnsureRoom(room?)` to make it work with context?

packages/react/etc/components-react.api.md
```diff
@@ -695,6 +710,20 @@ export function useSortedParticipants(participants: Array<Participant>): Partici
 // @public
 export const useSpeakingParticipants: () => Participant[];
 
+// @alpha
+export function useStartAudio({ room, props }: UseStartAudioProps): {
```
lukasIO, line 714: same here, moving `ensureRoom` from the component to the hook seems sensible to me

lukasIO (approved):
nice!

PR #489, opened 2023-05-22
Stringify room options

### LINKED ISSUE #488: Using options in LiveKitRoom component breaks app

\@livekit/components-react

### What I'm expecting
I want to be able to set the connection options in LiveKitRoom react component without facing any issues

### What happens instead
When connecting, I face the following error:

Screenshot 2023-05-22 at 13 11 35

### Reproduction

1 - Use any setup with LiveKitRoom react component

2 - Add options prop to that component. Here's mine (using the simple example in a fresh git clone):

```
<LiveKitRoom
  token={token}
  serverUrl={process.env.NEXT_PUBLIC_LK_SERVER_URL}
  connect={connect}
  onConnected={() => setIsConnected(true)}
  onDisconnected={handleDisconnect}
  audio={true}
  video={true}
  // HERE vvvv
  options={{
    adaptiveStream: true,
    dynacast: true,
    stopLocalTrackOnUnpublish: true,
    videoCaptureDefaults: {
      resolution: VideoPresets.h360.resolution,
    },
  }}
  ////////////////////
>

...
```

3 - Start server and connect

4 - Profit

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 13.3.1
    CPU: (8) arm64 Apple M2
    Memory: 113.72 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.20.0 - ~/.nvm/versions/node/v16.20.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.20.0/bin/yarn
    npm: 8.19.4 - ~/.nvm/versions/node/v16.20.0/bin/npm
  Browsers:
    Chrome: 113.0.5672.126
    Safari: 16.4
```

### Severity

serious, but I can work around it

### Additional Information

_No response_

lukasIO:
fixes #488

### DIFF
--- a/packages/react/src/components/LiveKitRoom.tsx
+++ b/packages/react/src/components/LiveKitRoom.tsx
@@ -120,7 +120,7 @@ export function useLiveKitRoom(props: LiveKitRoomProps) {
 
   React.useEffect(() => {
     setRoom(passedRoom ?? new Room(options));
-  }, [options, passedRoom]);
+  }, [JSON.stringify(options), passedRoom]);
 
   const htmlProps = React.useMemo(() => mergeProps(rest, setupLiveKitRoom()), [rest]);
 
@@ -186,7 +186,15 @@ export function useLiveKitRoom(props: LiveKitRoomProps) {
       log.debug('disconnecting because connect is false');
       room.disconnect();
     }
-  }, [connect, token, connectOptions, room, onError, serverUrl, simulateParticipants]);
+  }, [
+    connect,
+    token,
+    JSON.stringify(connectOptions),
+    room,
+    onError,
+    serverUrl,
+    simulateParticipants,
+  ]);
 
   React.useEffect(() => {
     if (!room) return;
--- a/.changeset/sweet-gorillas-whisper.md
+++ b/.changeset/sweet-gorillas-whisper.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-react": patch
+---
+
+Stringify room options


packages/react/src/components/LiveKitRoom.tsx
```diff
@@ -120,7 +120,7 @@ export function useLiveKitRoom(props: LiveKitRoomProps) {
 
   React.useEffect(() => {
     setRoom(passedRoom ?? new Room(options));
-  }, [options, passedRoom]);
+  }, [JSON.stringify(options), passedRoom]);
```
lukasIO, line 123: actually not sure if this is ideal. we do have options that are probably not intended to be stringified... 
e.g. `expWebAudioMix: { audioContext: new AudioContext }`...

Ocupe: Have you tried passing it the `new AudioContext`? Is it possible that `JSON.stringify` would throw an exception in this case?

Maybe it would be better if we wrap the options internally in a useState hook? Downside, you can only set the options once.

lukasIO: audio context (as an example) gets stringified to `{}` which would be fine. Still slightly concerned about other options potentially having an impact here, but will have to look more closely through what we currently offer.

lukasIO: checked again, stringifying should be fine!

PR #491, opened 2023-05-25
Correct Export Of LayoutContextType

richthegxd:


Ocupe:
Thank you 👍

lukasIO:
@Ocupe we'll need to add a changeset for this so that the updated package gets released

PR #493, opened 2023-05-25
fix: export of LayoutContextType

Ocupe:
Rexport `LayoutContextType` explitly as type to work with tsconfig `isolatedModules`.


Ocupe:
I was to slow :D. Already fixed by community ❤️ https://github.com/livekit/components-js/pull/491

PR #496, opened 2023-05-26
Adding a resize listener for the useSize hook

richthegxd:
Hello everyone, in my app and your your app (https://meet.livekit.io) there was a bug in which the **--lk-max-visible-tiles variable** was constantly changing its value due to which there was such a defect in the **CarouselLayout** component (attached video).

As I understood it, this was due to the multiple creation of **ResizeObserver**.

I checked my solution and this bug doesn't happen with it anymore. Also, the useGridLayout component, which uses the **useSize** hook, normally calculates variables.

Can you please see my solution, or we can try to solve this problem together with ResizeObserver.

Video example:
https://www.awesomescreenshot.com/video/17758562?key=7761dd3662adb0bd3d78493c377678a0

### DIFF
--- a/packages/react/src/hooks/internal/useResizeObserver.ts
+++ b/packages/react/src/hooks/internal/useResizeObserver.ts
@@ -100,18 +100,19 @@ export type UseResizeObserverCallback = (
 
 export const useSize = (target: React.RefObject<HTMLDivElement>) => {
   const [size, setSize] = React.useState({ width: 0, height: 0 });
-  React.useLayoutEffect(() => {
+
+  const handleResize = () => {
     if (target.current) {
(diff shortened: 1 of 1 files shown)


lukasIO:
thanks for the PR! 
the behaviour you're observing is probably due to the scrollbar popping up and changing the size of the element. 
we thought we had fixed that already, but apparently it is still happening in some cases. cc @Ocupe 

listening "only" to resize events will not be enough, because that will not trigger when the element itself (and not the window) changes its size. 

I'll close this PR, as it's not fixing what we want to fix (which is the flapping behaviour when the scrollbar becomes visible) and open an issue instead.

PR #501, opened 2023-05-30
Catch publish promises when first connecting room

lukasIO:


packages/react/src/components/LiveKitRoom.tsx
```diff
@@ -128,15 +128,16 @@ export function useLiveKitRoom(props: LiveKitRoomProps) {
...
+
+      log.debug('trying to publish local tracks');
+      Promise.all([
+        localP.setMicrophoneEnabled(!!audio, typeof audio !== 'boolean' ? audio : undefined),
+        localP.setCameraEnabled(!!video, typeof video !== 'boolean' ? video : undefined),
+        localP.setScreenShareEnabled(!!screen, typeof screen !== 'boolean' ? screen : undefined),
+      ]).catch((e) => {
         log.warn(e);
         onError?.(e as Error);
-      }
```
Ocupe, line 139: Was the error not caught because the function calls are asynchronous and not awaited?

Ocupe (approved):
Looks good, if the error is now handelt. 👍

packages/react/src/components/LiveKitRoom.tsx
lukasIO: yup, that was the bug!

Issue #515, opened 2023-06-05
Data messages without a topic should not show up in the chat


lukasIO:
### Select which package(s) are affected

\@livekit/components-core, \@livekit/components-react

### What I'm expecting
unrelated data messages should not show up in the chat.

### What actually happens
It seems like data messages without a topic are generating empty ghost messages in the chat component

### Reproduction

User reported the issue on slack https://livekit-users.slack.com/archives/C01KVTJH6BX/p1685575390807229?thread_ts=1685575350.386999&cid=C01KVTJH6BX

>  I'm sending data channel messages (without a topic if that matters because of the above issue I'm having), and any time I send one, it results in an empty chat message being added to the chat prefab looking like this:

### System Info

```shell
-
```

### Severity

serious, but I can work around it

PR #516, opened 2023-06-05
Only react to topics defined in the message observable

lukasIO:
fixes https://github.com/livekit/components-js/issues/515


Ocupe (approved):
👍

PR #518, opened 2023-06-07
update generated icons

Ocupe:
When running `yarn build` the generated icons update from doing imports like this:
`import { SVGProps } from 'react';` to doing it like that:
`import type { SVGProps } from 'react';`


lukasIO (commented):
finally it's supported ootb 🎉

lukasIO (approved):

PR #521, opened 2023-06-09
add setLogLevel function

Ocupe:
Add a setLogLevel function. By default this function sets the log level for both `@livekit/components-react` and the `@livekit-client` package. But there is an option to set `@livekit-client` independently.

### DIFF
--- a/packages/core/src/logger.ts
+++ b/packages/core/src/logger.ts
@@ -14,15 +14,7 @@ type SetLogLevelOptions = {
  * To set the `@livekit-client` log independently, use the `liveKitClientLogLevel` prop on the `options` object.
  * @public
  */
-export function setLogLevel(
-  level: LogLevel,
-  options: SetLogLevelOptions = { liveKitClientLogLevel: undefined },
-): void {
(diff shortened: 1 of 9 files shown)


packages/core/src/logger.ts
```diff
@@ -1,6 +1,28 @@
...
+};
+
+/**
+ * Set the log level for both the `@livekit/components-react` package and the `@livekit-client` package.
+ * To set the `@livekit-client` log independently, use the `liveKitClientLogLevel` prop on the `options` object.
+ * @public
+ */
+export function setLogLevel(
+  level: LogLevel,
+  options: SetLogLevelOptions = { liveKitClientLogLevel: undefined },
```
lukasIO, line 19: nit: `liveKitClientLogLevel` could simply default to `level` instead of `undefined`, which makes the if clause redundant

lukasIO (approved):

packages/core/src/logger.ts
Ocupe: Good catch!

PR #522, opened 2023-06-09
Play an empty track at all time to ensure autoplay on safari

lukasIO:
depends on https://github.com/livekit/client-sdk-js/pull/733

The current implementation requires an update to `getEmptyAudioTrack` that sets its volume to 0, otherwise users would constantly hear the sine wave playing.

### DIFF
--- a/packages/react/src/components/controls/StartAudio.tsx
+++ b/packages/react/src/components/controls/StartAudio.tsx
@@ -1,5 +1,5 @@
-import { setupStartAudio } from '@livekit/components-core';
-import type { Room } from 'livekit-client';
+import { setupStartAudio, log } from '@livekit/components-core';
+import { getEmptyAudioStreamTrack, type Room } from 'livekit-client';
 import * as React from 'react';
 import { useEnsureRoom, useRoomContext } from '../../context';
 import { useObservableState } from '../../hooks/internal/useObservableState';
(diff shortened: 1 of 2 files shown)


packages/react/src/components/controls/StartAudio.tsx
```diff
@@ -44,6 +46,20 @@ export function useStartAudio({ room, props }: UseStartAudioProps) {
...
+  log.warn(`canPlayAudio: ${canPlayAudio}`);
+
+  const dummyAudio = React.useRef<HTMLAudioElement>();
+
+  React.useEffect(() => {
+    if (canPlayAudio && !dummyAudio.current) {
+      dummyAudio.current = document.createElement('audio');
+      dummyAudio.current.autoplay = true;
+      //@ts-ignore
+      dummyAudio.current.playsInline = true;
```
Ocupe, line 58: Is there a reason why `playsInline` is set? It seems to be just a "video" tag attribute 🤔.

Ocupe (approved):
👍

packages/react/src/components/controls/StartAudio.tsx
lukasIO: ah, yeah, that makes sense, no reason. will remove it

lukasIO:
instead of handling this on the react level we could also do this
1. in the core package with `document.append(emptyAudioPlayer)` and call play on `startAudio`
2. move the logic for it into client-sdk-js so that all implementations get it. we could also make it Safari only for now

lukasIO:
closing in favor of https://github.com/livekit/client-sdk-js/pull/733

PR #523, opened 2023-06-12
Docs gen v3

Ocupe:
This PR:
* adds the `tooling/api-documenter` package, which converts `*.api.json` files into markdown files.
* updates the `docs-gen` package to use `api-extractor` and `api-documenter` to generate the docs.
* removes the unneeded `props-docs` package
* Add the `docs/docs` package which contains the generated markdown files. The package is intended to be consumed by the docs page.
* Add the api-extractor workflow to the core package, as it already exists in the react package.
* Add some package and trubo scripts to easily generate the docs (`yarn gen:docs` and `yarn gen:docs:watch`).


lukasIO (approved):
lgtm

PR #525, opened 2023-06-15
add screen share feature detection

Ocupe:
This PR replaces the condition wheater to show the screen sharing button in the control bar. The current detection was based on is-mobile-browser, which did not work for iPads. The new detection is based on feature-detection whether screen sharing is possible or not.

Thanks to Adam Conway for reporting this 👍

### DIFF
--- a/packages/core/src/helper/featureDetection.ts
+++ b/packages/core/src/helper/featureDetection.ts
@@ -1,0 +1,6 @@
+/**
+ * Returns `true` if the browser supports screen sharing.
+ */
+export function supportsScreenSharing(): boolean {
+  return navigator?.mediaDevices && !!navigator.mediaDevices.getDisplayMedia;
+}
--- a/packages/core/src/helper/featureDetection.ts
+++ b/packages/core/src/helper/featureDetection.ts
@@ -2,5 +2,9 @@
  * Returns `true` if the browser supports screen sharing.
  */
 export function supportsScreenSharing(): boolean {
-  return navigator?.mediaDevices && !!navigator.mediaDevices.getDisplayMedia;
+  return (
+    typeof navigator !== 'undefined' &&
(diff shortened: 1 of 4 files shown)


packages/core/src/helper/featureDetection.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Returns `true` if the browser supports screen sharing.
+ */
+export function supportsScreenSharing(): boolean {
+  return navigator?.mediaDevices && !!navigator.mediaDevices.getDisplayMedia;
```
lukasIO, line 5: haven't tested `navigator?.mediadevices` on non browser envs, if you have, feel free to ignore the suggestion
```suggestion
  return typeof navigator !== 'undefined' && navigator.mediaDevices && !!navigator.mediaDevices.getDisplayMedia;
```

packages/react/src/prefabs/ControlBar.tsx
```diff
@@ -80,7 +80,7 @@ export function ControlBar({ variation, controls, ...props }: ControlBarProps) {
     [variation],
   );
 
-  const isMobile = React.useMemo(() => isMobileBrowser(), []);
+  const browserSupportsScreenSharing = React.useMemo(() => supportsScreenSharing(), []);
```
lukasIO, line 83: do we need a memo for this?

lukasIO (approved):
way cleaner!

packages/react/src/prefabs/ControlBar.tsx
Ocupe: probably not

packages/core/src/helper/featureDetection.ts
Ocupe: Good call, better safe than sorry!

PR #527, opened 2023-06-19
Add facingMode detection

Ocupe:
Take advantage of the new facingMode detection https://github.com/livekit/client-sdk-js/pull/738 to stop mirroring the local participant's camera track when using backward facing mobile or OBS virtual cameras.

### DIFF
--- a/packages/react/etc/components-react.api.md
+++ b/packages/react/etc/components-react.api.md
@@ -458,7 +458,7 @@ export function useClearPinButton(props: ClearPinButtonProps): {
 
 // @public (undocumented)
 export function useConnectionQualityIndicator(options?: ConnectionQualityIndicatorOptions): {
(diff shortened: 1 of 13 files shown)


packages/react/etc/components-react.api.md
```diff
@@ -458,7 +458,7 @@ export function useClearPinButton(props: ClearPinButtonProps): {
...
+    className: any;
```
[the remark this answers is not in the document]
Ocupe: probably typescript missmatch, its gone now

packages/react/src/hooks/useFacingMode.ts
```diff
@@ -0,0 +1,35 @@
...
+  trackReference: TrackReferenceOrPlaceholder,
+): 'user' | 'environment' | 'left' | 'right' | 'undefined' {
+  if (!trackReference.participant.isLocal) {
+    return 'undefined';
+  }
+
+  let mediaStreamTrack: LocalTrack | undefined;
+  if (trackReference.publication === undefined) {
+    if (trackReference.participant.videoTracks.size > 0) {
+      mediaStreamTrack = Array.from(trackReference.participant.videoTracks.values())[0]
```
lukasIO, line 22: what's the idea here?

Ocupe: If you are alone in the room the the publication piped into the hook is unknown. This is away to still get the local track.

packages/react/src/hooks/useFacingMode.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (!trackReference.participant.isLocal) {
+    return 'undefined';
+  }
+
+  let mediaStreamTrack: LocalTrack | undefined;
+  if (trackReference.publication !== undefined) {
+    mediaStreamTrack = trackReference.publication.track as LocalTrack;
+  }
+
+  if (mediaStreamTrack) {
```
lukasIO, line 24: nit: the three conditions here could be boiled down to
```
if (ref.publication instanceof LocalTrackPublication)
...
else 
return undefined
```

lukasIO (approved):

PR #529, opened 2023-06-21
Update active device selection

Ocupe:
This PR uses the new client-sdk-js API (https://github.com/livekit/client-sdk-js/pull/743) to get the current selected device. This allows switching media devices even if there is no local track.

- To use this feature I had to update the requirements for the package `livekit-client` to version `1.11.2`.
  - !Updating the package resulted in some new ts errors that I have fixed or ts-ignored!

### DIFF
--- a/packages/core/src/components/mediaDeviceSelect.ts
+++ b/packages/core/src/components/mediaDeviceSelect.ts
@@ -1,9 +1,8 @@
-import type { LocalAudioTrack, LocalVideoTrack, Room } from 'livekit-client';
-import { Track } from 'livekit-client';
-import { BehaviorSubject, map, mergeWith } from 'rxjs';
+import type { Room } from 'livekit-client';
+import { BehaviorSubject, mergeWith } from 'rxjs';
 import { log } from '../logger';
-import { observeParticipantMedia } from '../observables/participant';
 import { prefixClass } from '../styles-interface';
+import { createActiveDeviceObservable } from '../observables/room';
 
 export type SetMediaDeviceOptions = {
   /**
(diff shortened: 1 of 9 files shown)


packages/core/src/components/mediaDeviceSelect.ts
```diff
@@ -17,42 +16,17 @@ export function setupDeviceSelector(kind: MediaDeviceKind, room?: Room) {
...
-              break;
-            default:
-              localTrack = undefined;
-              break;
-          }
-          return localTrack?.mediaStreamTrack.getSettings()?.deviceId;
-        }),
-        mergeWith(activeDeviceSubject),
-      )
+    ? createActiveDeviceObservable(room, kind).pipe(mergeWith(activeDeviceSubject))
```
lukasIO, line 19: do we actually need the merge with activeDeviceSubject if the room is available?

packages/core/src/components/mediaDeviceSelect.ts
```diff
@@ -17,42 +16,17 @@ export function setupDeviceSelector(kind: MediaDeviceKind, room?: Room) {
...
-          ?.track?.getDeviceId();
-      } else if (kind === 'audioinput') {
-        actualDeviceId = await room.localParticipant
-          .getTrack(Track.Source.Microphone)
-          ?.track?.getDeviceId();
-      }
+
       if (actualDeviceId !== id && id !== 'default') {
-        log.warn(`Failed to select the desired device. Desired: ${id}. Actual: ${actualDeviceId}`);
+        actualDeviceId = room.getActiveDevice(kind);
```
lukasIO, line 29: should we have a warn/info log still if the check fails of actualDeviceId against the one that was passed in by the user?

Ocupe: Sounds good!

lukasIO (approved):
🙏

PR #533, opened 2023-06-27
Combined Device Permissions in useMediaAllDevicesSelect

toger5:
depends on: https://github.com/livekit/client-sdk-js/pull/755
Add an additional useMediaDevices hook.
Next to useMediaDeviceSelect there is also useMediaAllDevicesSelect that asks for permissions for all devices simultaneously and makes the one selected by the user the active ones.

Overview:
 - `useMediaAllDevicesSelect`: tracks the state of all media device selections and updates the device selection in an active livekit room.
 - `useMediaDeviceSelect`: tracks one device and updates the active device in a livekit room.
 - `useMediaDevices`: only returns the device list.

All of them allow to also ask for permission (Which is mandatory if one wants to see the actual device labels and not only the id's)
Signed-off-by: Timo K <toger5@hotmail.de>

### DIFF
--- a/packages/core/src/observables/participant.ts
+++ b/packages/core/src/observables/participant.ts
@@ -126,9 +126,11 @@ export function participantEventSelector<T extends ParticipantEvent>(
       // @ts-ignore
       subscribe.next(params);
     };
+    // @ts-ignore
     participant.on(event, update);
 
     const unsubscribe = () => {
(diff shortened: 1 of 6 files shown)


toger5:
close in favor of the now `usePreviewtracks` hook.

PR #535, opened 2023-06-29
Improve media device selection

lukasIO:
depends on https://github.com/livekit/client-sdk-js/pull/759 to get the correct default selection after publishing

### DIFF
--- a/packages/react/src/components/controls/MediaDeviceSelect.tsx
+++ b/packages/react/src/components/controls/MediaDeviceSelect.tsx
@@ -33,7 +33,6 @@ export function useMediaDeviceSelect({ kind, room }: UseMediaDeviceSelectProps) 
 
   React.useEffect(() => {
     const listener = activeDeviceObservable.subscribe((deviceId) => {
-      console.log('active device changed', deviceId);
       if (deviceId) setCurrentDeviceId(deviceId);
     });
     return () => {
@@ -109,8 +108,8 @@ export function MediaDeviceSelect({
         <li
           key={device.deviceId}
           id={device.deviceId}
-          data-lk-active={device.deviceId.includes(activeDeviceId)}
-          aria-selected={device.deviceId.includes(activeDeviceId)}
+          data-lk-active={device.deviceId === activeDeviceId}
+          aria-selected={device.deviceId === activeDeviceId}
           role="option"
         >
           <button className="lk-button" onClick={() => handleActiveDeviceChange(device.deviceId)}>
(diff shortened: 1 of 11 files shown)


packages/react/src/components/controls/MediaDeviceSelect.tsx
```diff
@@ -87,9 +89,12 @@ export function MediaDeviceSelect({
     }
   }, [onDeviceListChange, devices]);
 
+  React.useEffect(() => {
+    onActiveDeviceChange?.(activeDeviceId);
+  }, [activeDeviceId]);
```
Ocupe, line 94: missing dep `onActiveDeviceChange`?

packages/react/src/components/controls/MediaDeviceSelect.tsx
```diff
@@ -87,9 +89,12 @@ export function MediaDeviceSelect({
...
+
   const handleActiveDeviceChange = async (deviceId: string) => {
-    setActiveMediaDevice(deviceId);
-    onActiveDeviceChange?.(deviceId);
+    await setActiveMediaDevice(deviceId, { exact: exactMatch });
   };
```
Ocupe, line 98: Maybe consider inlining?

packages/react/src/components/controls/MediaDeviceSelect.tsx
```diff
@@ -49,6 +49,7 @@ export interface MediaDeviceSelectProps extends React.HTMLAttributes<HTMLUListEl
   onActiveDeviceChange?: (deviceId: string) => void;
   onDeviceListChange?: (devices: MediaDeviceInfo[]) => void;
   initialSelection?: string;
+  exactMatch?: boolean;
```
Ocupe, line 52: Maybe a short docstring to explain what exactMatch does and what happens if there is no exact match.

Ocupe (approved):
Only small comments.

turbo.json
```diff
@@ -36,6 +36,7 @@
     },
     "api-check": {
       "dependsOn": ["^build"]
-    }
+    },
+    "api-check:update": {}
```
Ocupe, line 40: I think this is missing 
```
"api-check:update": {
     "outputs": ["temp/*", "ect/*"]
}
```

lukasIO: what is the outputs section used for? 🤔

packages/react/src/components/controls/MediaDeviceSelect.tsx
lukasIO: it's now wrapped in a try/catch so leaving as is

turbo.json
lukasIO: I think that's only useful if some other task would depend on `update`, right? 
For that case I don't really see a scenario where we would want to keep that info cached, as a call to it means something went wrong/API changed

PR #537, opened 2023-06-30
Prejoin update

lukasIO:
main objectives:

- request device specific permissions (and both at once if audio and video are specified)
- on subsequent requests, make sure it's only for the request of that device type (so that you don't get a prompt for both again, just because you're switching mics)
- make sure enumeration permission requests are only acquired after the other permissions have been successfully acquired

this is done primarily for a smoother experience on FF, where permissions are granted on a per-device basis which could, among other things, lead to multiple permission requests popping up for the same media device kind (e.g. two pop ups for microphone)

### DIFF
--- a/examples/nextjs/pages/prejoin.tsx
+++ b/examples/nextjs/pages/prejoin.tsx
@@ -1,0 +1,38 @@
+import * as React from 'react';
+import { setLogLevel } from '@livekit/components-core';
+import { LiveKitRoom, PreJoin, useToken, VideoConference } from '@livekit/components-react';
+import type { NextPage } from 'next';
+
+const MinimalExample: NextPage = () => {
+  //   const params = typeof window !== 'undefined' ? new URLSearchParams(location.search) : null;
+  //   const roomName = params?.get('room') ?? 'test-room';
+  //   const userIdentity = params?.get('user') ?? 'test-identity';
+  setLogLevel('debug', { liveKitClientLogLevel: 'warn' });
+  //   const [controlBarOptions, setControlBarOptions] = React.useState<ControlBarControls>({
+  //     chat: true,
+  //   });
+
+  //   const token = useToken(process.env.NEXT_PUBLIC_LK_TOKEN_ENDPOINT, roomName, {
+  //     userInfo: {
+  //       identity: userIdentity,
+  //       name: userIdentity,
+  //     },
+  //   });
+
+  //   React.useEffect(() => {
+  //     import('@livekit/track-processors').then((pkg) => {
+  //       setControlBarOptions({
+  //         camera: { processors: { 'Background Blur': pkg.BackgroundBlur(10) } },
+  //         chat: true,
+  //       });
+  //     });
+  //   }, []);
+
+  return (
+    <div data-lk-theme="default" style={{ height: '100vh' }}>
+      <PreJoin />
+    </div>
+  );
+};
+
+export default MinimalExample;
--- a/examples/nextjs/pages/prejoin.tsx
+++ b/examples/nextjs/pages/prejoin.tsx
@@ -3,7 +3,7 @@ import { setLogLevel } from '@livekit/components-core';
 import { LiveKitRoom, PreJoin, useToken, VideoConference } from '@livekit/components-react';
 import type { NextPage } from 'next';
 
-const MinimalExample: NextPage = () => {
+const PreJoinExample: NextPage = () => {
   //   const params = typeof window !== 'undefined' ? new URLSearchParams(location.search) : null;
   //   const roomName = params?.get('room') ?? 'test-room';
   //   const userIdentity = params?.get('user') ?? 'test-identity';
--- a/examples/nextjs/pages/prejoin.tsx
+++ b/examples/nextjs/pages/prejoin.tsx
@@ -1,32 +1,10 @@
 import * as React from 'react';
 import { setLogLevel } from '@livekit/components-core';
-import { LiveKitRoom, PreJoin, useToken, VideoConference } from '@livekit/components-react';
+import { PreJoin } from '@livekit/components-react';
 import type { NextPage } from 'next';
 
 const PreJoinExample: NextPage = () => {
-  //   const params = typeof window !== 'undefined' ? new URLSearchParams(location.search) : null;
-  //   const roomName = params?.get('room') ?? 'test-room';
-  //   const userIdentity = params?.get('user') ?? 'test-identity';
   setLogLevel('debug', { liveKitClientLogLevel: 'warn' });
-  //   const [controlBarOptions, setControlBarOptions] = React.useState<ControlBarControls>({
-  //     chat: true,
-  //   });
-
-  //   const token = useToken(process.env.NEXT_PUBLIC_LK_TOKEN_ENDPOINT, roomName, {
-  //     userInfo: {
-  //       identity: userIdentity,
-  //       name: userIdentity,
-  //     },
-  //   });
-
-  //   React.useEffect(() => {
-  //     import('@livekit/track-processors').then((pkg) => {
-  //       setControlBarOptions({
-  //         camera: { processors: { 'Background Blur': pkg.BackgroundBlur(10) } },
-  //         chat: true,
-  //       });
-  //     });
-  //   }, []);
 
   return (
     <div data-lk-theme="default" style={{ height: '100vh' }}>
@@ -35,4 +13,4 @@ const PreJoinExample: NextPage = () => {
   );
 };
 
-export default MinimalExample;
+export default PreJoinExample;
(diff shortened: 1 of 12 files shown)


packages/core/src/components/mediaDeviceSelect.ts
```diff
@@ -12,7 +12,11 @@ export type SetMediaDeviceOptions = {
   exact?: boolean;
 };
 
-export function setupDeviceSelector(kind: MediaDeviceKind, room?: Room) {
+export function setupDeviceSelector(
+  kind: MediaDeviceKind,
+  room?: Room,
+  localTrack?: LocalAudioTrack | LocalVideoTrack,
```
lukasIO, line 18: by allowing a track to be passed in, we can switch devices on that track, which allows us to use client-sdk-js for the switching logic here as well

packages/react/src/prefabs/PreJoin.tsx
```diff
@@ -87,31 +121,33 @@ export function usePreviewDevice<T extends LocalVideoTrack | LocalAudioTrack>(
...
-  }, [enabled, localTrack, deviceError]);
+  }, [enabled, localTrack, deviceError, isCreatingTrack]);
 
   // switch camera device
   React.useEffect(() => {
+    if (!localTrack) {
+      return;
+    }
     if (!enabled) {
       if (localTrack) {
```
toger5, line 146: this is redundant with the previous return on `!localTrack` ?

packages/react/src/prefabs/PreJoin.tsx
```diff
@@ -87,31 +121,33 @@ export function usePreviewDevice<T extends LocalVideoTrack | LocalAudioTrack>(
...
         log.debug(`muting ${kind} track`);
         localTrack.mute().then(() => log.debug(localTrack.mediaStreamTrack));
       }
-      return;
-    }
-    if (
+    } else if (
       localTrack &&
```
toger5, line 151: this is redundant as well

packages/react/src/prefabs/PreJoin.tsx
```diff
@@ -177,29 +215,42 @@ export const PreJoin = ({
...
+  const [audioDeviceId, setAudioDeviceId] = React.useState<string>(initialAudioDeviceId);
 
-  const video = usePreviewDevice(videoEnabled, videoDeviceId, 'videoinput');
+  const tracks = usePreviewTracks(
+    {
+      audio: audioEnabled ? { deviceId: initialAudioDeviceId } : false,
+      video: videoEnabled ? { deviceId: initialVideoDeviceId } : false,
+    },
+    onError,
+  );
```
toger5, line 233: I dont seem to find where the tracks are updated if `audioDeviceId`/`videoDeviceId` change?
Previously (with `usePreviewDevice` this happened when changing the params of the hook
```ts
const video = usePreviewDevice(videoEnabled, /*HERE:*/ videoDeviceId, 'videoinput');
```

toger5: Oh. found it. the tracks are now passed to the setupDeviceSelector... That is very reasonable. (Because it mimics the api which is present if a room is available)

packages/react/src/prefabs/PreJoin.tsx
lukasIO: true! thanks

packages/react/src/prefabs/PreJoin.tsx
lukasIO: yeah exactly. the idea is that `usePreviewTracks` is only used for the initial value in order to facilitate a unified permission dialog if both video and microphone are enabled. Subsequent permission requests are then handled on the track level directly. 
I hope this accounts for all edge cases, but it will definitely need some more testing/iterations to be sure

packages/react/src/components/controls/MediaDeviceSelect.tsx
```diff
@@ -16,19 +16,20 @@ export function useMediaDevices({ kind }: { kind: MediaDeviceKind }) {
...
   room?: Room;
+  track?: LocalAudioTrack | LocalVideoTrack;
 }
 
 /** @public */
-export function useMediaDeviceSelect({ kind, room }: UseMediaDeviceSelectProps) {
+export function useMediaDeviceSelect({ kind, room, track }: UseMediaDeviceSelectProps) {
   const roomContext = useMaybeRoomContext();
   // List of all devices.
   const deviceObserver = React.useMemo(() => createMediaDeviceObserver(kind), [kind]);
```
toger5, line 26: in `createMediaDeviceObserver` there still is a call to `getUserMedia` with just `audio:true, video: false` (or vice versa). This coul still resutl in duplicated/simultaneous calls to `getUserMedia`.
They should be skipeed if `usePreviewTracks` is used. (except id the js-client-sdk is able to properly skip those duplicated calls. also accounting for similar constraints liek `{video: true, audio:true}` `{video: {deviceId:"123"}, audio:{deviceId:234}}`

lukasIO: that's true, however for the `DeviceMenu` component this is solved, by only rendering the `MediaDeviceSelect` once the track is available, see https://github.com/livekit/components-js/pull/537/files#diff-5a2258ada71877932aa61092f670ecc56b371aab617a35c00f7bf68cc54e6ab2R96

lukasIO: once that track is enabled the `getUserMedia` call in room.getLocalDevices would only happen if the first call to `enumerateDevices` only returns dummy data. If however, permissions for camera and/or microphone have been granted already, it will skip the `getUserMedia` call altogether. Device Ids should not be impacting this at this point

toger5: That makes sense. I am sceptical however that this is obvious to sdk users who dont use the DeviceMenu component. I could see how adding `props.disabled` until the tracks have triggered the `getUserMedia` permissions is missed and as a result the user could be faced with two prompts (if a browser that does not handle this gracefully is used)

I could see that adding an option to `useMediaDeviceSelect` and `createMediaDeviceObserver` like `requestPermissions` could make this more transparent.
```ts
 /**
   * @requestPermissions this will call getUserMedia if the permissions are not yet given to enumerate the devices 
   * with device labels see: https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/enumerateDevices
   * in some browsers multiple calls to getUserMedia result in multiple permission prompts. It is adviced to 
   * disable requestPermissions until getUserMedia is called once with the desired devices outside of this hook.
   */
```

lukasIO: Agree that making this dependent on `disabled` isn't the most obvious of choices. 

I like the idea to make that available additionally! Thanks a lot for the documentation text, copied 90% of that over :)

Ocupe (approved):

examples/nextjs/pages/prejoin.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { setLogLevel } from '@livekit/components-core';
+import { LiveKitRoom, PreJoin, useToken, VideoConference } from '@livekit/components-react';
+import type { NextPage } from 'next';
+
+const MinimalExample: NextPage = () => {
+  //   const params = typeof window !== 'undefined' ? new URLSearchParams(location.search) : null;
+  //   const roomName = params?.get('room') ?? 'test-room';
+  //   const userIdentity = params?.get('user') ?? 'test-identity';
+  setLogLevel('debug', { liveKitClientLogLevel: 'warn' });
+  //   const [controlBarOptions, setControlBarOptions] = React.useState<ControlBarControls>({
```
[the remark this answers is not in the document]
lukasIO: no, will remove, thanks!

packages/react/src/components/controls/MediaDeviceSelect.tsx
```diff
@@ -16,23 +16,43 @@ export function useMediaDevices({ kind }: { kind: MediaDeviceKind }) {
...
   kind: MediaDeviceKind;
   room?: Room;
+  track?: LocalAudioTrack | LocalVideoTrack;
+  /**
+   * this will call getUserMedia if the permissions are not yet given to enumerate the devices with device labels.
+   * in some browsers multiple calls to getUserMedia result in multiple permission prompts.
+   * It's generally advised only flip this to true, once a (preview) track has been acquired successfully with the
+   * appropriate permissions.
+   *
+   * @see [MediaDeviceMenu](../../prefabs/MediaDeviceMenu.tsx)
```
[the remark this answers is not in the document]
lukasIO: thanks, that makes a lot of sense! 
just unfortunate that the link doesn't seem to get picked up by vscode in that case...

PR #549, opened 2023-07-07
Display active state of chat toggle

lukasIO:


packages/react/src/components/controls/ChatToggle.tsx
```diff
@@ -8,18 +8,23 @@ interface UseToggleChatProps {
...
+        {
+          className,
+          onClick: () => {
+            if (dispatch) dispatch({ msg: 'toggle_chat' });
+          },
         },
-      }),
-    [props, className, dispatch],
+        // @ts-expect-error custom data props
+        { 'data-lk-active': state?.showChat ? 'true' : 'false' },
```
Ocupe, line 25: Any reasons for the new custom data attribute? Have you considered using the `aria-ressed` attribute like we do for the other control buttons?

packages/styles/scss/components/controls/_button.scss
```diff
@@ -25,6 +24,11 @@
     }
   }
 
+  &:not([disabled])[data-active='true'] {
+    color: var(--accent-fg);
+    background-color: var(--accent-bg);
```
Ocupe, line 29: I think it is a bit unexpected that this does not use `var(--control-active-bg)`. If we want to keep the blue accent color (I don't mind if we use the default colors control btn colors), we should probably overwrite it in `_chat-toggle.scss`. 
If we decide to use aria-pressed instead of `data-lk-active`, the whole rule might be redundant because of the rule on line 19.

packages/styles/scss/components/controls/_chat-toggle.scss
```diff
@@ -1,3 +1,6 @@
+@use 'button';
+
 .chat-toggle {
+  @extend .button;
```
Ocupe, line 4: good call

packages/react/src/components/controls/ChatToggle.tsx
lukasIO: aria-pressed is a lot better, you're right!
We use `data-lk-active` in both `deviceSelector` and `paginationIndicator`. I guess at least for `deviceSelector` it should then also be using `aria-pressed` ?

PR #552, opened 2023-07-07
Add overload for  defaulting to undefined topic

lukasIO:


Ocupe (approved):
This hook is missing some documentation that explains what a topic is and what the expected hehavious is when you pass a topic to the hook.

Not a fan of function overloads, especially when used to handle different amounts and orders of arguments. I would rather see topic beeing non-optional and the default behind the curtain magic/behavior go away.

lukasIO:
the gain is 
```ts
useDataChannel(myCb)
```
vs. 
```ts
useDataChannel(undefined, myCb)
```

Ocupe:
> the gain is
> 
> ```ts
> useDataChannel(myCb)
> ```
> 
> vs.
> 
> ```ts
> useDataChannel(undefined, myCb)
> ```

I like the expressive and consistent nature of simply:
```tsx
useDataChannel('forced-to-always-pick-a-topic', myCallbackFn)
```
When I talked about the small gain, I was referring to "not having to type a topic string into the hook". In my opinion, if you have a default topic, you miss an opportunity to teach about the fact that you can have more than one data channel. But this is just a personal opinion.

PR #553, opened 2023-07-07
Only render layout components in browser (no SSR) in order to avoid u…

lukasIO:
…seLayoutEffect warnings


packages/core/src/helper/detectMobileBrowser.ts
```diff
@@ -1,4 +1,4 @@
-function isWeb(): boolean {
+export function isWeb(): boolean {
```
Ocupe, line 1: I would mark it with `/** @internal */` to remove it from the public API.

Ocupe (approved):
Not pretty, but I guess React forces us to :D

PR #559, opened 2023-07-12
Add data-lk-orientation for kind video

nwang92:
This handles cases where `source === Track.Source.Unknown` but still has `kind === 'video'` using the same logic as https://github.com/livekit/components-js/blob/main/packages/react/src/components/participant/ParticipantTile.tsx#L155

### DIFF
--- a/.changeset/seven-adults-try.md
+++ b/.changeset/seven-adults-try.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-react": patch
+---
+
+Add data-lk-orientation for all video tracks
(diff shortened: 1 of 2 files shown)


Ocupe (approved):
I think that makes sense. Thank you for your contribution!

lukasIO:
thanks for the PR! 
the difference to https://github.com/livekit/components-js/blob/main/packages/react/src/components/participant/ParticipantTile.tsx#L155 is that this referenced code is checking against `trackReference.source`. In the code here the check is against `publication.source`. 
I think your suggestion to check against `publication.kind` makes sense, but this makes the subsequent checks against `publication.source` redundant and we can remove them!

nwang92:
> thanks for the PR! the difference to https://github.com/livekit/components-js/blob/main/packages/react/src/components/participant/ParticipantTile.tsx#L155 is that this referenced code is checking against `trackReference.source`. In the code here the check is against `publication.source`. I think your suggestion to check against `publication.kind` makes sense, but this makes the subsequent checks against `publication.source` redundant and we can remove them!

Done for useMediaTrackBySourceOrName.ts. Did you want to change this for ParticipantTile.tsx as well or should I leave it as-is?

lukasIO (approved):
looks good now, thanks!

PR #560, opened 2023-07-13
Handle track references type change in updatePages

Ocupe:
This PR fixes the bug where e.g. a track reference changing from type `TrackReferencePlaceholder` to `TrackReference` was not picked up by the `updatePages` function.
Because we update based on ID, we did not pick up the change in type and stuck with the old track reference.


lukasIO (approved):
nice ❤️

PR #565, opened 2023-07-19
Move Hooks into hook dir

Ocupe:
This PR:

- Move component hooks to the hooks folder.
- fixes and simplifies some TS types
- exposes the `useFocusToggle` and `useToggleChat` hooks

### DIFF
--- a/packages/react/src/hooks/useToggleChat.ts
+++ b/packages/react/src/hooks/useToggleChat.ts
@@ -3,7 +3,7 @@ import { useLayoutContext } from '../context';
 import { mergeProps } from '../mergeProps';
 import * as React from 'react';
 
-interface UseToggleChatProps {
+export interface UseToggleChatProps {
   props: React.ButtonHTMLAttributes<HTMLButtonElement>;
 }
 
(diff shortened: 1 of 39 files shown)


packages/react/src/hooks/useToggleChat.ts
```diff
@@ -0,0 +1,32 @@
...
+import { mergeProps } from '../mergeProps';
+import * as React from 'react';
+
+/** @public */
+export interface UseToggleChatProps {
+  props: React.ButtonHTMLAttributes<HTMLButtonElement>;
+}
+
+/** @public */
+export function useToggleChat({ props }: UseToggleChatProps) {
```
lukasIO, line 12: we generally have the semantics of `use*Toggle` so I'd suggest renaming this to `useChatToggle` which aligns better with the component name, if we make this public

lukasIO (approved):

packages/react/src/hooks/useToggleChat.ts
Ocupe: good catch 👍

PR #566, opened 2023-07-23
Add MediaDevicesError event for use local participant

mpnri:
Hi.
#### Bug Description
Currently, if there is a problem with the microphone or camera after requesting access from the browser, the **component** using `useLocalParticipant` will not be aware of that and will not update the UI (until another event is called).

####  Fix
This change causes the `useLocalParticipant` hook to be updated in case of a problem accessing the microphone or camera (`MediaDevicesError` event) and deliver the relevant errors in the output.

### DIFF
--- a/.changeset/ninety-stingrays-join.md
+++ b/.changeset/ninety-stingrays-join.md
@@ -1,0 +1,6 @@
+---
+"@livekit/components-core": patch
+"@livekit/components-react": patch
+---
+
+Add MediaDevicesError event for use local participant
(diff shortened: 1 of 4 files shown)


lukasIO:
Hi! 
thanks for the PR first of all. 
We already have a `onMediaDeviceFailure` callback on the `LiveKitRoom` component.
Is that something that would work for your use case already?

mpnri:
Hmm, I think since this problem is related to re-rendering, using a context API or a state manager like redux could probably work with that callback.
However, since `@livekit/components-react` already uses context API, and the `useLocalParticipant` hook and the `ObserverParticipantMedia` function are responsible for providing data about the camera and microphone, such as if they are active or not, I thought it would be better to add `MediaDevicesError` event to the observer, and return the errors as react states in `useLocalParticipant`. This also makes the **component** (that uses the hook) re-render when we get an error while creating a local track.

lukasIO:
make sense to me! 
@Ocupe any objections?

Ocupe:
> make sense to me! @Ocupe any objections?

Nope, looking good! Thank you @mpnri 🙏

Ocupe (approved):

PR #569, opened 2023-07-27
Remove unnecessary props from audio and video elements

mpnri:
Remove unnecessary props from audio and video elements in DOM by destructuring.
here is an example of an audio element in `livekit/examples/nextjs` which has unnecessarily attributes like `participant` or `publication`:

media-props

### DIFF
--- a/.changeset/brown-beers-share.md
+++ b/.changeset/brown-beers-share.md
@@ -1,0 +1,6 @@
+---
+"@livekit/components-core": patch
+"@livekit/components-react": patch
+---
+
+Remove unnecessary props from audio and video elements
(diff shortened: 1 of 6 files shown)


Ocupe:
@mpnri thank you for this PR. Looking good 👍 
To ensure that the tests pass, you must update the `components-react.api.md` file. You can do this by simply running `yarn api:update` and then commiting the changes.

mpnri:
Done.
I was updating `components-react.api.md` and I saw that the `track` attribute was still in the DOM element. I looked at the `tracks.ts` file and I noticed that this attribute was added to the `trackReferences` list, even though it was not part of the `TrackReference` type. So I thought it was safe to remove it from the code.

lukasIO:
> So I thought it was safe to remove it from the code.

That has wider implications. The track needs to stay in there :)

lukasIO:
> even though it was not part of the TrackReference type

hah, I thought we still needed it, but I guess you're right. Thanks for spotting this!

lukasIO (approved):

PR #581, opened 2023-08-02
Expose custom message encoder/decoder from video conference

jmoguilevsky:
This change was done so that applications that use a different message serialization protocol, in our case it's protobuff, can customize how to parse/unparse the input sent in the message content for the chat

### DIFF
--- a/.changeset/long-geese-suffer.md
+++ b/.changeset/long-geese-suffer.md
@@ -1,0 +1,6 @@
+---
+'@livekit/components-core': patch
+'@livekit/components-react': patch
+---
+
+Expose custom message encoder/decoder from video conference


packages/core/etc/components-core.api.md
```diff
@@ -347,7 +347,7 @@ export type SetMediaDeviceOptions = {
 };
 
 // @public (undocumented)
-export function setupChat(room: Room): {
+export function setupChat(room: Room, messageEncoder?: (message: string, timestamp: number) => Uint8Array, messageDecoder?: (message: Uint8Array) => ChatMessage): {
```
lukasIO, line 350: I'd prefer to pack those two into an (optional) options object together

packages/core/src/components/chat.ts
```diff
@@ -41,9 +52,11 @@ export function setupChat(room: Room) {
 
   const isSending$ = new BehaviorSubject<boolean>(false);
 
+  const finalMessageEncoder = messageEncoder ? messageEncoder : encode;
```
lukasIO, line 55: nit: 
```suggestion
  const finalMessageEncoder = messageEncoder ?? encode;
```

lukasIO (commented):
thanks for the PR! 
I think that's a neat idea. Just some small comments from me.

packages/react/src/prefabs/Chat.tsx
```diff
@@ -4,25 +4,27 @@ import * as React from 'react';
...
 /** @public */
 export interface ChatProps extends React.HTMLAttributes<HTMLDivElement> {
   messageFormatter?: MessageFormatter;
+  messageEncoder?: MessageEncoder;
+  messageDecoder?: MessageDecoder;
 }
 
 /** @public */
-export function useChat() {
+export function useChat(messageEncoder?: MessageEncoder, messageDecoder?: MessageDecoder) {
```
lukasIO, line 20: same here for the options object

packages/core/etc/components-core.api.md
jmoguilevsky: That's perfect!
I've made the update

packages/core/src/components/chat.ts
```diff
@@ -16,7 +16,18 @@ export interface ReceivedChatMessage extends ChatMessage {
...
-export function setupChat(room: Room) {
+const encode = (message: string, timestamp: number) =>
+  encoder.encode(JSON.stringify({ message, timestamp }));
+
+const decode = (message: Uint8Array) => JSON.parse(decoder.decode(message)) as ChatMessage;
+
+export function setupChat(
+  room: Room,
+  options: {
+    messageEncoder?: (message: string, timestamp: number) => Uint8Array;
```
lukasIO, line 27: a couple more nits: 
could you make this accept a `ChatMessage` object directly?

packages/react/src/prefabs/Chat.tsx
```diff
@@ -4,25 +4,30 @@ import * as React from 'react';
...
 /** @public */
 export interface ChatProps extends React.HTMLAttributes<HTMLDivElement> {
   messageFormatter?: MessageFormatter;
+  messageEncoder?: MessageEncoder;
+  messageDecoder?: MessageDecoder;
 }
 
 /** @public */
-export function useChat() {
+export function useChat(options: {
```
lukasIO, line 20: the options object needs to be optional

jmoguilevsky: I think fallbacking to empty object is better, so I can still do 
```const { messageDecoder, messageEncoder } = options;```

File: packages/core/src/components/chat.ts [hunk 6]
```diff
@@ -16,7 +16,18 @@ export interface ReceivedChatMessage extends ChatMessage {
...
+    messageDecoder?: (message: Uint8Array) => ChatMessage;
```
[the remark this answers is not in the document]
jmoguilevsky: Right, because of the `from`, shouldn't the encoder have that information too?

packages/react/src/prefabs/Chat.tsx
lukasIO: you can do the fallback in `core`, but in the react package that unfortunately causes unneeded re-evaluations because (for some reason) default parameters in react like objects and arrays are always considered as "new" objects. 

you'll also want to include `options` in the useEffect dependency array

packages/core/src/components/chat.ts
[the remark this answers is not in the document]
lukasIO: No, the `from` gets added automatically when receiving the message

packages/react/src/prefabs/Chat.tsx
jmoguilevsky: Hi @lukasIO I've updated all comments can you recheck the PR?

packages/react/src/prefabs/Chat.tsx
```diff
@@ -45,7 +50,10 @@ export function useChat() {
 export function Chat({ messageFormatter, ...props }: ChatProps) {
   const inputRef = React.useRef<HTMLInputElement>(null);
   const ulRef = React.useRef<HTMLUListElement>(null);
-  const { send, chatMessages, isSending } = useChat();
+  const { send, chatMessages, isSending } = useChat({
```
lukasIO, line 53: This is causing an infinite render loop due to how the object is created. I'd suggest to do something like this instead
```ts
 const chatOptions = React.useMemo(() => {
    return { messageDecoder, messageEncoder };
  }, [messageDecoder, messageEncoder]);
  const { send, chatMessages, isSending } = useChat(chatOptions);
```

jmoguilevsky: Thanks, I updated the code with that update

lukasIO: Sorry, I should have been clearer on this: the reason why I suggested the `{messageDecoder, messageEncoder}` to be referenced directly in the `useMemo` call, is because the `...props` are further down spread on a `div` element. 
so it makes more sense to do 
```
export function Chat({ messageFormatter, messageDecoder, messageEncoder, ...props }: ChatProps) {
```

jmoguilevsky: Ah okay, I fixed that by just adding them in the Chat initial destructuring

lukasIO (approved):
looks good to me now, thanks for your patience with this!

lukasIO:
one last ask, could you run
```
yarn changeset
```

to add a changeset file. A patch release for `react` and `core` should be auto-selected and that's fine. 
As a description, the title of the PR works. 

Tried to add it myself, but didn't have the permissions to do it on your branch.

jmoguilevsky:
That's great!

I've run the changeset and set it as a patch update

lukasIO:
thanks, you'll need to commit that file so that the github bot picks it up!

PR #586, opened 2023-08-09
Use relative path for outputs

lukasIO:


lukasIO:
yeah sorry, it was a warning being logged when running a command

Issue #592, opened 2023-08-17
SyntaxError: Named export 'useDebounce' not found. [usehooks-ts]


sawariz0r:
### Select which package(s) are affected

\@livekit/components-react

### Describe the bug

Started getting this in the latest version (1.1.1), had to roll back to 1.1.0.
Seems to be an issue with the dependency `usehooks-ts`.

### Reproduction

Installing and running with 1.1.1 seems to cause the issue. 
Reverting to 1.1.0 solves the issue.

### Logs

```shell
file:///Users/[user]/[folder]/[project]/[project-package]/node_modules/@livekit/components-react/dist/index.mjs:1966
import { useDebounce, useIntersectionObserver } from "usehooks-ts";
         ^^^^^^^^^^^
SyntaxError: Named export 'useDebounce' not found. The requested module 'usehooks-ts' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'usehooks-ts';
const { useDebounce, useIntersectionObserver } = pkg;
```

### System Info

```shell
System:
    OS: macOS 13.3.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 122.53 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.11.0 - /opt/homebrew/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 8.19.2 - /opt/homebrew/bin/npm
    pnpm: 7.14.0 - /opt/homebrew/bin/pnpm
    Watchman: 2022.10.24.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 115.0.5790.170
    Safari: 16.4
  npmPackages:
    livekit-client: ^1.6.2 => 1.12.0 
    @livekit/components-react: ^1.0.7 => 1.1.1
```

### Severity

blocking all usage of LiveKit

lukasIO:
thanks for the report! which package manager are you using and which nextjs version (or in case it isn't nextjs, what bundler are you using?)

sawariz0r:
I'm using yarn primarily. Using next `13.1.4` :)

lukasIO:
thanks, I was able to reproduce it once but now it seems to work on subsequent installs/runs. 
[...]

PR #594, opened 2023-08-17
Make sure dependencies stay external for esm builds

lukasIO:
first attempt to solve #592 

the issue only manifested once the package was published, so guessing we'll have to do that again to verify if the fix actually works.

related issues:
https://github.com/juliencrn/usehooks-ts/issues/162
https://github.com/evanw/esbuild/issues/2164


Ocupe (approved):
🤞

PR #597, opened 2023-08-22
api documenter code gen update

Ocupe:
Some changes relate only to docs generation.


lukasIO:
which files have the non-auto generated changes?

Ocupe:
> which files have the non-auto generated changes?

image

PR #609, opened 2023-08-31
Use fenced code block as return value

Ocupe:
Switch from inline code to code blocks for better readability.

## Example

**Old**
`{     buttonProps: ClearPinButtonProps & {         className: string;         disabled: boolean;         onClick: () => void;     }; }`

**New**
```typescript
{
    buttonProps: ClearPinButtonProps & {
        className: string;
        disabled: boolean;
        onClick: () => void;
    };
}
```


tooling/api-documenter/src/documenters/MarkdownDocumenter.ts
Ocupe, line 1: This is the only manual code change, rest is code gen.

lukasIO (approved):

PR #618, opened 2023-09-01
fix handling of multiple tracks of the same source from the same participant

Ocupe:
This PR addresses a general problem with the way we distinguished tracks. In many places we used a combination of participant identity and track source to identify a track. This falls apart when you have participants with multiple tracks from the same source. For example, a robot participant with multiple camera tracks.

- To address the problem we now construct a `TrackReference` ID like this: `<participant_id>_<track_source>_<publication_id>`
- We also had to update the way `FocusToggle` and `useFocusToggle` work. We deprecated `participant` and `trackSource` parameters in favour of passing a `TrackReference` via the `trackRef` parameter directly.
- Some small changes to the visual stable update logic were also necessary. `TrackReferencePlaceholder` and `TrackReference` would produce the same ID to prevent tiles from changing position in the `GridLayout` or `Carousel`. Because we changed the way `TrackReference` IDs are constructed, we had to modify the existing update step to "migrate" `TrackReference` IDs from `TrackReferencePlaceholders` to `TrackReference`.

Special thanks to Sijmen Huizenga from the LiveKit community Slack for [reporting this bug.](https://livekit-users.slack.com/archives/C048FRL1N2C/p1693388065097569)

### DIFF
--- a/packages/react/src/components/TrackLoop.tsx
+++ b/packages/react/src/components/TrackLoop.tsx
@@ -1,8 +1,8 @@
 import type { TrackReference, TrackReferenceOrPlaceholder } from '@livekit/components-core';
-import { isTrackReference } from '@livekit/components-core';
 import * as React from 'react';
 import { TrackContext } from '../context/track-context';
 import { cloneSingleChild } from '../utils';
+import { getTrackReferenceId } from '@livekit/components-core';
 
 /** @public */
 export interface TrackLoopProps {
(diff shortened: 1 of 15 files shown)


packages/react/src/components/TrackLoop.tsx
```diff
@@ -31,14 +31,8 @@ export function TrackLoop({ tracks, ...props }: TrackLoopProps) {
...
       {tracks.map((trackReference) => {
-        const trackSource = isTrackReference(trackReference)
-          ? trackReference.publication.source
-          : trackReference.source;
         return (
-          <TrackContext.Provider
-            value={trackReference}
-            key={`${trackReference.participant.identity}_${trackSource}`}
-          >
+          <TrackContext.Provider value={trackReference} key={getTrackReferenceId(trackReference)}>
```
lukasIO, line 35: so much cleaner 😍

packages/react/src/components/participant/ParticipantTile.tsx
```diff
@@ -148,7 +151,7 @@ export function ParticipantTile({
             </div>
           </>
         )}
-        <FocusToggle trackSource={trackRef.source} />
+        <FocusToggle trackRef={trackRef} />
```
lukasIO, line 154: only wondering about naming here of `trackRef`. 

`track` is ambiguous, but are we using `trackRef` as a property somewhere else already?

lukasIO (approved):
nice! 
following semver somewhat, we could think about making it a minor release for components-react, as it's deprecating public APIs

PR #619, opened 2023-09-03
Only unsusbscribe if intersection is still false

lukasIO:
This fixes a bug where tiles coming into view might be unsubscribed from after ~3s because a debounce handler was kicking in that is supposed to make sure that we only unsubscribe from tracks after a certain timeout. 
However that also led to the effect that if the tiles were changing back into view within those 3s they would still get unsubscribed. Fix is simply to check on both the debounced and non-debounced intersection handler whether or not the element is currently intersecting.

### DIFF
--- a/.changeset/few-gifts-dress.md
+++ b/.changeset/few-gifts-dress.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-react": patch
+---
+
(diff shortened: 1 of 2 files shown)


davidzhao (approved):
lgtm

PR #622, opened 2023-09-05
Update AudioTrack and VideoTrack components to accept track references

Ocupe:


lukasIO (approved):
lgtm! 
a `minor` changeset would be good for this, too!

PR #627, opened 2023-09-07
Use `trackRef` as primary way of passing track info around, deprecate participant/source based APIs

lukasIO:
This reverts commit 6b4a4b6ec077eaf7f3875fb3c6c20e09e00cd9b2.


Ocupe (approved):
⏩

PR #632, opened 2023-09-11
make source and participant optional

Ocupe:
Make source and participant optional, so users don't have to provide them if they already pass a `trackRef'.

### DIFF
--- a/packages/react/src/hooks/useParticipantTile.ts
+++ b/packages/react/src/hooks/useParticipantTile.ts
@@ -19,9 +19,9 @@ export interface UseParticipantTileProps<T extends HTMLElement> extends React.HT
   onParticipantClick?: (event: ParticipantClickEvent) => void;
   htmlProps: React.HTMLAttributes<T>;
   /** @deprecated This parameter will be removed in a future version use `trackRef` instead. */
-  source: Track.Source;
+  source?: Track.Source;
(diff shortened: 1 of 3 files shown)


packages/react/src/hooks/useParticipantTile.ts
```diff
@@ -39,10 +39,16 @@ export function useParticipantTile<T extends HTMLElement>({
   const maybeTrackRef = useMaybeTrackRefContext();
   const p = useEnsureParticipant(participant);
   const trackReference = React.useMemo(() => {
+    const _source = trackRef?.source ?? maybeTrackRef?.source ?? source;
+    if (_source === undefined) {
+      throw new Error(
+        'Missing track `source`, provided it via `trackRef`, `source` property or via `TrackRefContext`.',
```
lukasIO, line 45: just a thought: with all the deprecation code we're currently writing, it might already be beneficial to have a custom error for this in order to not have to write out this whole string all the time.

lukasIO (approved):
thanks for catching this!

packages/react/src/hooks/useParticipantTile.ts
Ocupe: Good idea. Expanding on that, we could have some predefined errors for missing contexts with helpful error messages and how to fix the problem. Maybe even a link to the docs.

PR #639, opened 2023-09-18
Add note about deprecated APIs in the generated documents

Ocupe:
This PR adds deprecation warnings for top-level APIs such as components and hooks, as well as parameter-specific warnings.

> **Warning**
> Everything under `docs/docs/generated-docs` is automatically generated and does not need to be reviewed.

SCR-20230918-mxoa
SCR-20230918-mxrs


lukasIO:
lgtm. 
I guess we wouldn't want to remove deprecated docs entirely?
Keeping deprecated one's around is nice of course, but also a lot more to sift through for new users...

lukasIO:
Definitely worth thinking about, I guess they would still be part of the in-line-js-docs anyways. Just not publicly featured on the website.

PR #641, opened 2023-09-19
Fix start audio component flash issue

harrisonlo:
Fix start audio component flash issue with canPlayAudio set to true by default until detected otherwise

### DIFF
--- a/.changeset/light-spies-brake.md
+++ b/.changeset/light-spies-brake.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-react": patch
+---
+
+Set the default value for `canPlayAudio` returned by the `useStartAudio` hook to `true` to avoid flashing issues.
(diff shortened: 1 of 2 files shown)


Ocupe (approved):
@harrisonlo thanks for the PR. I think this is a reasonable change, especially because it fixes your problem. 👍
@lukasIO any objections?

davidzhao:
I think this aligns with the client-sdk-js defaults as well: https://github.com/livekit/client-sdk-js#audio-playback

```
LiveKit takes an optimistic approach so it's possible for this value to change from true to false when we encounter a browser error.
```

Ocupe:
> I think this aligns with the client-sdk-js defaults as well: https://github.com/livekit/client-sdk-js#audio-playback
> 
> ```
> LiveKit takes an optimistic approach so it's possible for this value to change from true to false when we encounter a browser error.
> ```

Great! Thank you for the context 👍

lukasIO:
ah, sorry, only saw this now. What we would want is to set the init value to `room.canPlaybackAudio` instead of a hard coded value.

PR #647, opened 2023-09-23
Remove non-published orphaned icons 

hehehai:
Duplicated with the `SvgQualityUnknownIcon.tsx` file?


Ocupe:
@hehehai Thanks for bringing this to our attention 👍. 
I have made some additions. All icons are generated, so I made sure we remove all of them before generating new ones. I also added a warning that the icons are generated automatically.

hehehai:
@Ocupe thx! Cool, automatically generating resources is a good way to synchronize, and I found this problem when I copied resource files.

I'm trying to write vue components, But I see that the component library seems to be refactoring.
https://github.com/livekit/components-js/compare/main...hehehai:components-js:hehehai/support-vue

lukasIO (approved):

PR #682, opened 2023-11-01
Update PreJoin description

Ocupe:
This PR updates the docstring on PreJoin to indicate that it should not be nested within a LiveKitRoom component. The problem is the TrackToggle component, which is used both inside the LiveKitRoom context (TrackToggleButton) and outside it (PreJoin). When the TrackToggle is nested inside a PreJoin and a LiveKitRoom component, it can't handle this and mistakenly thinks it should work as if it were inside a room context. 

Short-term workaround: Indicate that the PreJoin should not be used inside a RoomContext.
Long term solution: Don't use the TrackToggle inside the PreJoin. Probably better to have a separate and specific implementation for the PreJoin component.


packages/react/src/prefabs/PreJoin.tsx
```diff
@@ -202,7 +202,7 @@ export function usePreviewDevice<T extends LocalVideoTrack | LocalAudioTrack>(
  * On submit the user decisions are returned, which can then be passed on to the `LiveKitRoom` so that the user enters the room with the correct media devices.
  *
  * @remarks
- * This component is independent from the `LiveKitRoom` component and don't has to be nested inside it.
+ * This component is independent from the `LiveKitRoom` component and should not be nested inside it.
```
lukasIO, line 205: nit:
```suggestion
 * This component is independent from the `LiveKitRoom` component and should not be nested inside of it.
```

lukasIO (approved):

PR #683, opened 2023-11-03
Add usePersistentUserChoices hook to save device settings and username to local storage

Ocupe:
This pull request adds a new hook, `usePersistentUserChoices`, which provides functionality to save and load user choices. This will allow users to persist their device settings and username across sessions.

The user choices object stored in local storage looks like this:
```
export type UserChoices = {
  videoInputEnabled: boolean;
  audioInputEnabled: boolean;
  videoInputDeviceId: string;
  audioInputDeviceId: string;
  username: string;
};
```

Settings in local storage are updated by changes in the `ControlBar` and the `PreJoin` component. The general goal is to always use the media devices that were last used.
I have added functionality to prevent the `usePersistentUserChoices` from writing to local storage, but by default we write to local storage.


lukasIO:
I like the idea, I'm wondering if the stuff that is in `core` right now, might actually make sense in `livekit-client` directly? That way not only react users would profit.

packages/core/src/persistent-storage/user-choices.ts
```diff
@@ -0,0 +1,92 @@
...
+ * Saves user choices to local storage.
+ * @param deviceSettings - The device settings to be stored.
+ * @alpha
+ */
+export function saveUserChoices(
+  deviceSettings: UserChoices,
+  /**
+   * Whether to prevent saving user choices to local storage.
+   */
+  preventSave: boolean = false,
```
lukasIO, line 56: could we instead of the boolean parameter simply _not_ call the function if the user doesn't intend to save?

packages/core/src/persistent-storage/user-choices.ts
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * @alpha
+ */
+export function loadUserChoices(
+  defaults?: Partial<UserChoices>,
+  /**
+   * Whether to prevent loading from local storage and return default values instead.
+   * @defaultValue false
+   */
+  preventLoad: boolean = false,
```
lukasIO, line 77: same question as above

packages/react/src/hooks/usePersistentUserChoices.ts
```diff
@@ -0,0 +1,60 @@
...
+export interface UsePersistentUserChoicesOptions {
+  /**
+   * The default value to use if reading from local storage returns no results or fails.
+   */
+  defaults?: Partial<UserChoices>;
+  /**
+   * Whether to prevent saving to persistent storage.
+   * @defaultValue false
+   */
+  preventSave?: boolean;
```
lukasIO, line 18: wondering if `save` and `load` would make more sense than negating the option with `prevent*`

Ocupe: I have the feeling that `preventSave=true` tells you a bit more than just `save=true`.

packages/core/src/persistent-storage/user-choices.ts
Ocupe: I wanted the default `defaultUserChoices` (if no defaults are passed by the user) to be defined and handled within the core. This API evolved from that decision. We could move the same logic one layer up into components-react, but that would increase the API surface a bit, and (if we support more frameworks in the future) we would have to reimplement it.

packages/react/src/hooks/usePersistentUserChoices.ts
Ocupe: What about `skipSave: boolean`?

lukasIO: Really just wanted to hear your opinion on it, I think your argument is fair. Let's stick with `prevent*` then!

packages/core/src/persistent-storage/user-choices.ts
lukasIO: got it, that makes sense to me!

lukasIO (approved):

PR #691, opened 2023-11-06
In the examples, add a helper function to generate a random user identity

Ocupe:
This PR is intended to improve the user experience when using the examples. It is the direct result of a user request that encountered this problem.

When running the examples and then navigating directly to `localhost:3000/minimal` with two tabs, you end up with two identical users kicking each other out of the room. 
To prevent this from ever happening again, we now generate a random user id if no user id is set with the `user` query parameter.

### DIFF
--- a/.changeset/tame-nails-deny.md
+++ b/.changeset/tame-nails-deny.md
@@ -1,0 +1,5 @@
+---
+"@livekit/component-example-next": patch
+---
+
+Add helper function to generate random user identity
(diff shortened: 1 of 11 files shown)


lukasIO:
nice, while we're at it, can we change it in a way that clicking the link in the overview menu also uses this function? so that opening the link multiple times generates a new user identity each time?

Ocupe:
> nice, while we're at it, can we change it in a way that clicking the link in the overview menu also uses this function? so that opening the link multiple times generates a new user identity each time?  

I have removed the Faker dependency and only use the new random user ID function.

lukasIO:
huh, that's even better :)

PR #694, opened 2023-11-07
Update USER_CHOICES_KEY to reflect its purpose

Ocupe:
This pull request updates the `USER_CHOICES_KEY` constant.

### DIFF
--- a/.changeset/great-needles-relate.md
+++ b/.changeset/great-needles-relate.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-core": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/great-needles-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+"@livekit/components-core": patch
+---
+
+Update USER_CHOICES_KEY to reflect its purpose.
```
lukasIO, line 5: I think we haven't released the local storage stuff yet, so we don't really need a changeset here, might even be confusing

lukasIO (approved):

PR #695, opened 2023-11-07
Add saveUsername and make PreJoin backwards compatible

Ocupe:
This pull request adds a new function, `saveUsername`, to the `usePersistentUserChoices` hook. This function allows the user's chosen username to be saved to persistent storage.

For backward compatibility and simplicity, the properties of the UserChoices type have been updated to match the original LocalUserChoices type. To avoid TS-hell The goal now is to deprecate the `e2ee` and `sharedPassphrase` properties before changing the type to `UserChoices`.


lukasIO (approved):
lgtm!

PR #699, opened 2023-11-09
Auto publish docs with release action

lukasIO:
and remove generated docs from git


tooling/docs-gen/package.json
```diff
@@ -11,7 +11,7 @@
     "directory": "eslint-plugin-tsdoc"
   },
   "scripts": {
-    "gen:docs": "yarn gen:api-documenter && yarn gather-api-files && yarn format-gen-docs",
+    "gen:docs": "pnpm gather-api-files && pnpm gen:api-documenter && pnpm format-gen-docs",
```
lukasIO, line 14: @Ocupe it seems that gather-api-files should run first? I was getting an error when api-documenter runs first as the ./temp folder does not yet exist

Ocupe: Makes sens :D

Ocupe (approved):

PR #707, opened 2023-11-09
Specify TS paths directly to avoid losing declarations when building

lukasIO:


lukasIO:
@Ocupe any idea why api-check is complaining about so many (seemingly) new things now? Does the action fail on `warnings` or must there be something else going on?

lukasIO:
this has caused some other issues, albeit still unclear why this only surfaced now, see https://github.com/livekit/components-js/issues/1154

Issue #603, opened 2023-08-28
Support 2 chats in one room


ochurae:
### Select which package(s) are affected

\@livekit/components-core

### Describe the bug

I need to show 2 separate chats on a page in the same livekit room.

### What I'm expecting
When I message in one of them, i want to see my message in both chats.

### What happens instead
When I message in one chat, it doesn't share that message with other one.

I'm using "useChat" hook in both chat components to interact with livekit api.
After some investigations I found a function "setupChat", that creates a new chat setup for every "useChat" hook usage, and passes a message of localParticipant to "messageSubject" after it was successfully sent.
"messageSubject" is scoped in "setupChat" function, so can't share messages of localParticipants with other chats in the livekit room.

I think that solution here could be a creation of one more rxjs Subject, that would be sharing a messages across the room.
Ex.:
```
  const send = async (message: string) => {
    const timestamp = Date.now();
    const encodedMsg = finalMessageEncoder({ message, timestamp });
    isSending$.next(true);
    try {
      await sendMessage(room.localParticipant, encodedMsg, DataTopic.CHAT, {
        kind: DataPacket_Kind.RELIABLE,
      });
-> sharedMessageSubject.next({ <-
        payload: encodedMsg,
        topic: DataTopic.CHAT,
        from: room.localParticipant,
      });
    } finally {
      isSending$.next(false);
    }
  };
```

Maybe you have better idea how to solve this issue?

### Reproduction

Create 2 chats components using "useChat" hook in one livekit room -> type in one chat and send message -> you see that, chatMessages updated with what you've typed -> check other chat data, there will be no update of chatMessages

### System Info

```shell
System:
    OS: macOS 13.4.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 11.14 MB / 16.00 GB
  Binaries:
    Node: 18.16.1 - ~/.nvm/versions/node/v18.16.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.16.1/bin/yarn
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.1/bin/npm
  Browsers:
    Chrome: 116.0.5845.110
```

### Severity

serious, but I can work around it

PR #713, opened 2023-11-20
Add support for multiple chat windows

lukasIO:
Closes https://github.com/livekit/components-js/issues/603

### DIFF
--- a/packages/core/src/components/chat.ts
+++ b/packages/core/src/components/chat.ts
@@ -19,34 +19,42 @@ export interface ReceivedChatMessage extends ChatMessage {
 export type MessageEncoder = (message: ChatMessage) => Uint8Array;
 /** @public */
 export type MessageDecoder = (message: Uint8Array) => ReceivedChatMessage;
+/** @public */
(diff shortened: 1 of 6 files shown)


packages/core/src/components/chat.ts
```diff
@@ -19,34 +19,42 @@ export interface ReceivedChatMessage extends ChatMessage {
...
+type RawMessage = {
+  payload: Uint8Array;
+  topic: string | undefined;
+  from: Participant | undefined;
+};
 
 const encoder = new TextEncoder();
 const decoder = new TextDecoder();
 
+const topicSubjectMap: Map<string, Subject<RawMessage>> = new Map();
```
lukasIO, line 38: @Ocupe currently, this gets never cleared/disposed. We won't want to merge it as is. 
I was thinking about when the best way to dispose of specific entries would be. 
My thoughts:
- do some ref counting and dispose when refCount === 0
- only dispose on room disconnect (we'd have to be sure that we catch that event)

Ocupe (approved):

PR #720, opened 2023-11-30
Emit missing style warning on VideoConference when in development mode

lukasIO:


Ocupe (commented):
How about creating a custom hook instead of a function call in a useEffect hook? Then we can drop it into other prefabs with minimal effort and lines of code.

`useWarnAboutMissingStyles()` or something

lukasIO:
@Ocupe addressed your comments

Ocupe (approved):
Nice 👍

PR #728, opened 2023-12-05
Add setLogExtension to core

mpnri:
Add the `setLogExtension` function for setting extension for both `livekit-client`, and `@livekit/components-core`: just like the `setLogLevel` function.

### DIFF
--- a/.changeset/curly-papayas-bow.md
+++ b/.changeset/curly-papayas-bow.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-core": patch
+---
+
+Add setLogExtension to core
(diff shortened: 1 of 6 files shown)


lukasIO:
this is a nice addition, could you add a changeset and also expose the function from `components-react` directly (re-export in packages/react/src/index`)

mpnri:
> this is a nice addition, could you add a changeset and also expose the function from `components-react` directly (re-export in packages/react/src/index`)

Done👍.

lukasIO (approved):
nice, thank you!

PR #729, opened 2023-12-05
Add useParticipantTracks hook

lukasIO:


packages/react/src/hooks/useParticipantTracks.ts
```diff
@@ -0,0 +1,25 @@
...
+import { useObservableState } from './internal';
+import type { Participant, Track } from 'livekit-client';
+import { useMaybeParticipantContext } from '../context';
+
+/**
+ * @public
+ */
+export function useParticipantTracks(
+  sources: Track.Source[],
+  participant?: Participant,
```
Ocupe, line 13: I think accepting a participant.identity is closer to streaming use case and you don't have to get a participant first. The other way around, it is simple to get the identity if you already have a participant object.

packages/core/etc/components-core.api.md
```diff
@@ -319,6 +319,21 @@ export interface ParticipantMedia<T extends Participant = Participant> {
 // @public (undocumented)
 export function participantPermissionObserver(participant: Participant): Observable<ParticipantPermission | undefined>;
 
+// @public (undocumented)
+export const participantTrackEvents: ParticipantEvent[];
+
+// Warning: (ae-forgotten-export) The symbol "RequireAtLeastOne" needs to be exported by the entry point index.d.ts
+//
+// @public (undocumented)
+export type ParticipantTrackIdentifier = RequireAtLeastOne<{
```
Ocupe, line 328: Personally, I find this type rather confusing and not too easy to understand. I'm hesitant about making this part of the public API.

packages/react/src/hooks/useTrack.ts
```diff
@@ -9,6 +10,7 @@ export interface UseTrackOptions {
 }
 
 /** @public */
-export function useTrack(trackRef: TrackIdentifier, options: UseTrackOptions = {}) {
-  return useMediaTrackBySourceOrName(trackRef, options);
+export function useTrack(source: Track.Source, participant?: Participant) {
```
Ocupe, line 13: Non-deterministic if there are multiple tracks with the same source. You pointed out that this is an edge case or only for advanced users, but I'm not a fan of hiding the underlying problem for the sake of not returning an array of tracks. I think this problem applies to useTrack as well.

lukasIO: do you haven an alternative suggestion for using a single track?

packages/core/etc/components-core.api.md
lukasIO: we can get rid of it, if we introduce the `byName` hooks instead. Does that sound good to you?

packages/react/src/hooks/useTrack.ts
Ocupe: I think 
```tsx
const cameraTracks = useParticipantTracks([Track.Source.Camera], 'participant-id')

if(cameraTracks.length > 0){
  const cameraTrack = cameraTracks[0]
}
```

is the best we can currently offer. I think only if we have something like useTrackById could we return a single track.

lukasIO: I understand your point, but I would argue that users expect such an API as they can do the same on all the other client SDKs (i.e. `participant.getTrack(Track.SourceCamera)` and get a single track back). While I hear you saying that we don't have to replicate that behaviour, for me, it's more a question of how to minimise the mental overhead juggling different APIs/SDKs

Ocupe: I think most developers will only be exposed to one SDK, so I'm not sure if this is a big problem. On top of that, react components are the first of its kind (component package), so I don't think we have any need to replicate the APIs from the underlying package.
I'm not going to die on this hill, just to summarise things:
The hook is syntactic sugar to make an already simple thing even simpler. If there were no downside, I would be totally fine with it. I personally would not want to sacrifice correctness and predictability for a slight improvement in developer experience.

packages/core/etc/components-core.api.md
lukasIO: this type is now only exposed on the core, not on the reacts package

packages/react/src/hooks/useTrack.ts
lukasIO: removed `useTrack` for now in order to get a release out quickly. We can decide to re-introduce it later

packages/react/src/hooks/useParticipantTracks.ts
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import type { TrackReference } from '@livekit/components-core';
+import { participantTracksObservable } from '@livekit/components-core';
+import { useObservableState } from './internal';
+import type { Track } from 'livekit-client';
+import { useMaybeParticipantContext, useRoomContext } from '../context';
+
+/**
+ * @public
```
Ocupe, line 9: Missing some basic documentation with an example.

packages/react/src/hooks/useTrackByName.ts
```diff
@@ -1,15 +1,11 @@
-import { useEnsureTrackRef } from '../context';
-import type { UseMediaTrackOptions } from './useMediaTrack';
-import { useMediaTrackBySourceOrName } from './useMediaTrackBySourceOrName';
-import type { TrackReferenceOrPlaceholder } from '@livekit/components-core';
+import type { Participant } from 'livekit-client';
+import { useEnsureParticipant } from '../context';
+import { useTrackRefBySourceOrName } from './useTrackRefBySourceOrName';
 
 /**
  * @public
```
Ocupe, line 6: Docs :)

Ocupe (commented):
Some documentation, but otherwise looks good!

PR #734, opened 2023-12-14
Fix render loop in TrackToggle onChange

harrywebdev:
This fixes an infinite render loop introduced by #732 by not using memoized callbacks for TrackToggle's onChange prop.


lukasIO:
oh, thanks for the PR, yeah, we should fix this immediately. 
Can you just sign the CLA so that I can merge this?

harrywebdev:
Thanks! Also included a changeset, hopefully it's correct. Cheers!

lukasIO:
perfect, thank you!

Issue #665, opened 2023-10-23
Add a button to close the chat window 


alechaye:
### Describe the problem

If you open the chat window, the only way to close it is by pressing the same button that was used to open it. It is not intuitive for an end user. 
<img width="217" alt="image" src="https://github.com/livekit/components-js/assets/106412628/ef122e8e-769e-4cc0-9327-604dcec5d5f9">

### Describe the proposed solution

Adding a 'X' button or '->' to signal that you can close the chat by pressing there will greatly improve UI experience. 

### Alternatives considered

Adding the button.

### Importance

would make my life easier

lukasIO:
thanks for the report, we're aware of it, we forgot to release a new version of the styles package alongside the components-react package. 
We'll do that probably today!

PR #740, opened 2023-12-23
Added a button to close the chat window

govind-io:
resolves issue #665

The new chat window looks like:
image


lukasIO:
thanks for the PR. I circled this back to our design department. 
The Feedback is
> "Messages" should be centered horizontally in the top, not in the remaining space.
> Close button is more prominent than "Send", which hierarchically shouldn't be the case. I would try making it just the "X", and showing button bg on hover

govindgovind852:
I have made the below changes based on the feedback, please let me know if it looks fine or something else needs to be done:

1. Messages is centered aligned horizontally in the complete space, isntead of remaining space
2. Close button just shows "X" now untill hovered, when hovered it shows the background color.

Without hover on button:
image

With hover on the button:
image

Ocupe (approved):
Looks good to me! Many thanks for your contribution! 👍

PR #751, opened 2024-01-12
Pin all direct dependencies

lukasIO:


lukasIO:
curious why this change has an impact on package size 👀

davidzhao (approved):
NICE!

PR #753, opened 2024-01-13
Use room as dependency to recreate localParticipant observable

lukasIO:


davidzhao (approved):
thanks for the quick fix!

PR #755, opened 2024-01-15
Add possibility to edit/update chat messages

lukasIO:
shows a little `edit` button next to local messages when hovering.

edited messages get an `(edited)` text addition next to their timestamp when rendered.

### DIFF
--- a/.changeset/dull-dots-mate.md
+++ b/.changeset/dull-dots-mate.md
@@ -1,0 +1,5 @@
+---
+"@livekit/component-example-next": patch
+---
+
+Add possibility to edit/update chat messages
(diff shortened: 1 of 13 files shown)


Ocupe:
**Some comments:**
- Editing the message directly in the speech bubble is a rather uncommon way. Most messaging apps I'm aware of let you edit the message in the input file. So you perform an action to enter edit mode, and then you edit the message as you would when creating it in the first place. Once finished, you update the original message.
- Is it a good idea to be able to go into edit mode on more than one message at a time?
- How would I go into edit mode on mobile or on a platform where hovering is not supported?
- When editing a message, because the font size is less than 16px, the iPhone zooms in on the edit block. I think this is probably not what we want.
- Personally, I would remove the parentheses around **edited**.

lukasIO:
Valid points, I'll split this in two PRs, one for the hooks and one for the UI. As you didn't mention anything about the hooks, I'll assume those parts are alright with you

Ocupe:
I only had a quick look at the hooks, but they seemed fine!

Issue #759, opened 2024-01-31
useMediaDevice / useMediaDeviceSelect "can" return undefined


heilmela:
### Select which package(s) are affected

\@livekit/components-react

### Expected
useMediaDevice and relating hooks should not return nullables (default to empty array)
### Actual
Returns undefined when browser blocks device permission. 
This is at least an error in the typing, but seems to result in runtime errors aswell in livekit internals

### Reproduction

- Set mic/camera permissions to "Block" in the browser and call useMediaDevices
- Some livekit internals seem to expect non nullable devices aswell so we get a lot of random errors with long stack traces
- Happened after upgrade 1.3.0 -> 1.5.3

### Logs

```shell
Example Stack trace:

Unhandled Runtime Error
TypeError: Cannot read properties of undefined (reading 'find')
Call Stack
eval
javascript/node_modules/.pnpm/@livekit+components-react@1.5.3_livekit-client@1.15.10_react-dom@18.2.0_react@18.2.0_tslib@2.6.2/node_modules/@livekit/components-react/dist/index.mjs (3286:0)
commitHookEffectListMount
../node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (23150:0)
commitPassiveMountOnFiber
../node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (24926:0)
commitPassiveMountEffects_complete
../node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (24891:0)
commitPassiveMountEffects_begin
../node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (24878:0)
commitPassiveMountEffects
../node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (24866:0)
flushPassiveEffectsImpl
../node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (27039:0)
flushPassiveEffects
../node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (26984:0)
eval
[...]

PR #768, opened 2024-02-05
Do not crash in usePreviewDevice when user has not granted permission to media device

dnetteln:
When the user has not granted access to the media devices, the `usePreviewDevice` hook crashes on `devices.find()` This is particularly annoying because unhandled errors in hooks cannot be caught easily and will crash the whole application into the ErrorBoundary.

### DIFF
--- a/.changeset/four-yaks-grin.md
+++ b/.changeset/four-yaks-grin.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-react": patch
+---
+
+Do not crash in usePreviewDevice when user has not granted permission to media device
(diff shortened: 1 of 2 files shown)


lukasIO:
thanks for the PR! 

There's also a similar report for `useMediaDevice` [here](https://github.com/livekit/components-js/issues/759), would you want to add a fix for that as well?

dnetteln:
I had a quick look at it and this seems to be buried a bit deeper in the codebase and I don't have all the context how you handle Observables in this repo so I'll pass on that.

lukasIO:
that's understandable! Thanks for your contribution :)

lukasIO (approved):

PR #781, opened 2024-02-29
Add support for a settings modal render prop

lukasIO:


lukasIO:
hm. very unfortunate that adding this as an optional dependency seems to increase the LiveKitRoom component's size by 50%. No idea why that would be the case.

davidzhao (commented):
this is awesome!

question is should this be added to the prefab? or is it a Meet-only feature.. and how should we handle it if the filters package isn't installed? It seems like the option should be removed?

Ocupe (commented):
Very cool that we have noise cancellation! Eager to learn more about how it is implemented!
I'm not sure we want to take on the increase in API area. The settings menu seems to be something very specific in most applications. I think users will have specific designs and requirements that will rarely match something we could provide.
We had a similar discussion about whether encryption should have an end-user UI. I think it is a nice feature for us to be able to turn it on or off, but I assume for most production apps the decision whether or not to add noise cancellation is made by the developer of the app? So having a UI for it solves more our use case that the use case of our users? DZ's suggestion to move the settings menu to Meet seems reasonable to me, and/or turn it into an example for users to draw inspiration from.

If we decide that we want to add the settings prefab anyway, I noticed some styling bugs that we should take care of before merging this.
SCR-20240301-kudt
SCR-20240301-ktps

On a more general note. I'm not sure if we should increase the number of prefabs. From reports from the community, it seems that most people just use them as examples and not something to use as is. I'm more and more convinced that the prefabs should be high quality examples rather than something we include in the package (maybe with the exception of the videoconferencing component). I think our time is best spent focusing on providing the necessary building blocks to build them.

Ocupe (approved):

PR #782, opened 2024-02-29
Fix check for process in style warning

lukasIO:


Ocupe (commented):
How do I test this? If I remove the styles package from the nextjs example and then go to the minimal example, I don't get any warning.

lukasIO:
IIRC next used to inject `process.env` also for the browser. But it seems it doesn't do that anymore? 
Not sure. But would be good to get this in quickly nevertheless as it breaks production builds of components.

Ocupe (approved):

PR #798, opened 2024-03-05
Stop propagation of key events on chat input element

lukasIO:


Ocupe:
Do you have any context as to what motivated the change? 🙏

Ocupe (approved):

PR #801, opened 2024-03-11
Use blue accent colour for stop screen share button

### LINKED ISSUE #776: More prominent "Stop screen share" button
### Describe the problem

Maybe I'm getting old, but whenever I try to stop sharing my screen, I'm tempted to hit the big red "Leave" button instead:

image

(This has actually happened more than a few times now).

As a Firefox user, I occasionally hit its UI control for this:

image

However that thing is (a) relatively small, and (b) seems to pop up on a random monitor (I'm using multiple).

### Describe the proposed solution

Maybe add a little color to the "Stop screen share" button (`background-color: var(--lk-accent-bg)`):

image

and/or dim the "Leave" button while screen sharing is active:

image

### Alternatives considered

It would be nice to have a dedicated CSS variable for the button, as this is the minimal CSS selector needed to change the styling:

```css
/* apply accent color (blue) to screen share button, when screen sharing is active */
[data-lk-theme] .lk-button[data-lk-source="screen_share"][data-lk-enabled="true"]:not(:disabled) {
  background-color: var(--lk-accent-bg)
}
```

And even with a dedicated CSS variable, dimming the "Leave" button needs a bit more creativity:

```css
/* dim leave button, when screen sharing is active; should work even if chat is disabled */
[data-lk-theme] .lk-button[data-lk-source="screen_share"][data-lk-enabled="true"]:not(:disabled) + .lk-chat-toggle + .lk-disconnect-button, 
[data-lk-theme] .lk-button[data-lk-source="screen_share"][data-lk-enabled="true"]:not(:disabled)                   + .lk-disconnect-button {
  --lk-control-bg: var(--lk-control-active-bg);
}
```

### Importance

nice to have

### Additional Information

_No response_

lukasIO:
closes #776

### DIFF
--- a/.changeset/violet-pillows-draw.md
+++ b/.changeset/violet-pillows-draw.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-styles": patch
+---
+
+Use blue accent colour for stop screen share button
--- a/packages/styles/scss/components/controls/_button.scss
+++ b/packages/styles/scss/components/controls/_button.scss
@@ -25,6 +25,14 @@
     }
   }
 
+  /* apply accent color (blue) to screen share button, when screen sharing is active */
+  &[data-source='screen_share'][data-enabled='true'] {
+    background-color: var(--accent-bg);
+    &:hover {
+      background-color: var(--accent2);
+    }
+  }
+
   &:disabled {
     opacity: 0.5;
   }


Ocupe (approved):
💙

PR #819, opened 2024-04-02
Update useTracks.ts

FreitasAssis:
FIX error: Property 'getTrackPublications' does not exist on type 'Participant'.ts(2339)


davidzhao:
this is expected. Components have been updated to work with livekit-client 2.0. Please reference the migration guide for details: https://docs.livekit.io/guides/migrate-from-v1/

PR #826, opened 2024-04-10
Add react-native package

lukasIO:
this PR re-exports specific state handling and context hooks that can be used within react-native without further modification. 

There are some minor typing inconveniences, e.g. 
```tsx
function MyComponentsStage() {
  const tracks = useTracks([Track.Source.Camera]);

  console.log(tracks);

  return (
    <>
      {tracks.map((trackRef) => (
        <VideoView
          key={trackRef.participant.identity}
          videoTrack={trackRef.publication.track as VideoTrack | undefined}
        ></VideoView>
      ))}
    </>
  );
}
```

where we need to cast the publication.track as a VideoTrack right now as `useTracks` only returns general publications without being specific about the track type.

If we want to avoid people having to install yet another package, we can also think about doing the re-export of these hooks directly in the `@livekit/react-native` package that exists already today.


davidliu:
Will need to play more with it, but seems to work well. I think the only thing we'll want to add is a `LiveKitRoom` version for react native, since using `useLiveKitRoom` and setting up the context providers feels a little clunky.

lukasIO:
> I think the only thing we'll want to add is a LiveKitRoom version

totally agree! 

Would you guys prefer a separate package (the components-react-native one) or adding all these hooks to the already existing react-native package?

lukasIO:
closing in favor of https://github.com/livekit/client-sdk-react-native/pull/139

PR #827, opened 2024-04-12
Support referencing components via refs

lukasIO:
not being able to reference the underlying elements has been a long standing caveat with the react components.
This PR adds the ability to reference e.g. the underlying HTMLVideoElement of the `VideoTrack` component via

```ts
const videoEl = React.createRef(null);

return <VideoTrack ref={videoEl} ... />
```


davidzhao (approved):
lgtm!

PR #830, opened 2024-04-17
Improve track acquiring and handling in usePreviewTracks

lukasIO:
This PR tries to mitigate `NotReadableErrors` by making sure we only ever perform one `createLocalTracks` request at a time. 
Users reported seeing errors on Windows when trying to acquire specific cameras, so this is an attempt to fix that. 

depends on https://github.com/livekit/client-sdk-js/pull/1110 being released

### DIFF
--- a/packages/react/src/prefabs/PreJoin.tsx
+++ b/packages/react/src/prefabs/PreJoin.tsx
@@ -11,6 +11,7 @@ import {
   facingModeFromLocalTrack,
   Track,
   VideoPresets,
+  Mutex,
 } from 'livekit-client';
 import * as React from 'react';
 import { MediaDeviceMenu } from './MediaDeviceMenu';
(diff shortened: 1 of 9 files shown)


packages/react/src/utils.ts
```diff
@@ -71,3 +71,38 @@ export function useForwardedRef<T extends HTMLElement>(ref: React.ForwardedRef<T
 
   return innerRef;
 }
+
+export class Mutex {
```
davidzhao, line 75: should we export & use the version from client-sdk-js?

davidzhao (approved):

packages/react/src/prefabs/PreJoin.tsx
```diff
@@ -58,29 +59,33 @@ export function usePreviewTracks(
 ) {
   const [tracks, setTracks] = React.useState<LocalTrack[]>();
 
+  const trackLock = new Mutex();
```
lukasIO, line 62: this should be a memo

packages/react/src/utils.ts
lukasIO: makes sense, https://github.com/livekit/client-sdk-js/pull/1110

PR #834, opened 2024-04-19
Fix audio playing issue

rnakano:
It appears that there is a bug where the audio from `<AudioTrack>` is not playing during `<VideoConference>` in v2.1.0 and v2.1.1.
This likely involves the same issue as with the `<VideoTrack>` described in https://github.com/livekit/components-js/pull/831, where the refs aren't being passed correctly.
I have made the necessary corrections on the AudioTrack side.

Related slack threads: https://livekit-users.slack.com/archives/C01KVTJH6BX/p1713514083336509

### DIFF
--- a/.changeset/lemon-snakes-pull.md
+++ b/.changeset/lemon-snakes-pull.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-react": patch
+---
+
+Fix audio playing issue by using imperative handle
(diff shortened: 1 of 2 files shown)


lukasIO:
Sorry, only saw this right after I just merged https://github.com/livekit/components-js/pull/835. 
Thanks a lot for the report and fix!

rnakano:
No problem, thanks for the quick response!

PR #844, opened 2024-04-24
Fix releasing of local tracks in prejoin

lukasIO:
more subtleties when navigating away from prejoin. These changes make sure that the locally scoped variable is used to stop previously created tracks within the useEffect. 
Tested both against meet and the example and seems to consistently work now


Simba14 (approved):
Amazing, thanks for looking into this @lukasIO

davidzhao (approved):
awesome!

PR #847, opened 2024-04-24
Don't auto subscribe to all audio tracks automatically in RoomAudioRenderer

lukasIO:
for users with `autoSubscribe: false` automatically subscribing to all audio publications is rather unexpected


davidzhao (approved):
makes sense!

PR #853, opened 2024-04-26
Add hooks for transcriptions

lukasIO:
main changes with transcription related logic are 
- [`transcription.ts` in core](https://github.com/livekit/components-js/pull/853/files#diff-032aec1b73fad5993524f25e56e21aa41a52d952fb246f5274a3d0e9ae2aff55)
- [and `useTrackTranscription` in react](https://github.com/livekit/components-js/pull/853/files#diff-cc5e1a7177de5a5389f3a2f9ab59d0844c4efb9a91dec621f64399799f988625)


examples/nextjs/pages/prejoin.tsx
```diff
@@ -5,19 +5,23 @@ import { PreJoin, setLogLevel } from '@livekit/components-react';
...
-      <PreJoin
-        defaults={{ videoDeviceId: '' }}
-        onSubmit={(values) => {
-          values.audioDeviceId;
-        }}
-        onValidate={(values) => {
-          return true;
-        }}
-      />
+      <button onClick={() => setVisible((val) => !val)}>Toggle PreJoin</button>
```
davidzhao, line 13: this is part of the other PR right?

packages/react/src/hooks/useTranscription.ts
```diff
@@ -0,0 +1,25 @@
+import type { RemoteParticipant, RemoteTrackPublication, Transcription } from 'livekit-client';
+import { createTranscriptionObserver } from '@livekit/components-core';
+import * as React from 'react';
+import { useRoomContext } from '../context';
+
+/**
+ * @alpha
+ */
+export function useTranscription(
+  onTranscription: (
```
davidzhao, line 10: Does it make sense to wrap this? instead of having the user directly use the JS SDK event?

I'm thinking that might be more direct and requires less translation. Ideally the interface we offer in components contains the state of the last N transcriptions? like `useTranscriptions() -> []Transcription`, and similarly for participants: `useParticipantTranscriptions(p)`

and Transcription could be:

```
interface Transcription {
  participant,
  publication,
  segments,
```

davidzhao: this is just thinking out loud.. maybe each transcription should contain a single segment? so we can properly sort it by time. especially when metadata sync is in

examples/nextjs/pages/prejoin.tsx
lukasIO: It is unrelated to the main objective of the PR, but a useful addition in general to debug the prejoin component

packages/react/src/hooks/useTranscription.ts
lukasIO: yeah, tagging you for this PR was a bit premature, sorry about that. 
I'll play around with some designs on how to handle transcription state in conjunction with track syncs.

lukasIO:
@biglittlebigben  `trackSyncTimeObserver` is currently based on a `TrackSyncTime` event that we register for each remote track and that consistently emits new timing information. Currently it's reading out the rtp timestamp from the synchronization sources, see https://github.com/livekit/client-sdk-js/pull/1119/files#diff-d844533d54d1e1cedf8ed4b95a2eebeea4035b2ff72b21e89cb655edb7f0154fR85

davidzhao (approved):

davidzhao (commented):
nice!

PR #871, opened 2024-05-28
Support passing room to useIsEncrypted

renekliment:
I am aware this is alpha, so I am not sure how much support this gets. 

I would find it useful if the hook could also get the room passed directly as other hooks so it can be used in a place where `Room` is manually created, where the room context is not available. :pray: 

I was not sure whether the hook should only get `options` where both `participant` and `room` could be. I can rewrite it to that if that makes more sense. :pray:


lukasIO:
this makes sense, thanks for your contribution!

PR #874, opened 2024-05-29
Update focused placeholder to trackRef

### LINKED ISSUE #873: Bug in FocusLayout after logging in with camera turned off

\@livekit/components-react, examples

### Describe the bug

When a user enters a video conference and turns off the camera in the Prejoin component, the FocusLayout is on the user with the camera turned off, then if you try to turn on the camera in Video Conference component, this user goes to the CarouselLayout and the camera turns on there, and in place of the FocusLayout an empty black window appears with a duplicate name and microphone status of that same user. In this bug, I expect that when the camera is turned on, the user remains attached to the FocusLayout and does not go to the CarouselLayout.

**[Video of the bug](https://youtu.be/KE39bNoCvVQ)** - ( Youtube )

What is shown in the video in the official livekit demo, the same bug appears on my project.

### Reproduction

1. Prejoint Component
2. Turn off the video
3. Join video conference
4. Make a focus layout on yourself user
5. Turn on camera
6. Bug appears
7. User goes to carousel layout
8. Black window appears on focus layout

### Logs

_No response_

### System Info

```shell
System - macOS Sonoma (14.3.1). Browser - Google Chrome - Version 125.0.6422.77 (Official build), (arm64). 
Laptop - Macbook Air M1.
```

### Severity

blocking all usage of LiveKit

### Additional Information

I hope I showed how this bug appears in reality on the Livekit demo. I hope this problem will be solved and everything will work as it should.

lukasIO:
closes #873 

Previously, if a placeholder was set as a pinned track, the focused track would not update from a placeholder to a trackReference automatically, as the focused track would only have been stored once on the layout context. 
This is a workaround to ensure we update placeholders to trackRefs for the focusedTrack

### DIFF
--- a/.changeset/nervous-rocks-peel.md
+++ b/.changeset/nervous-rocks-peel.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-react": patch
+---
+
+Update focused placeholder to trackRef
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: pnpm/action-setup@v2
         with:
-          version: 8
+          version: 9
       - name: Use Node.js 20
         uses: actions/setup-node@v4
         with:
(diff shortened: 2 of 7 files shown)


Ocupe (approved):
👍

PR #893, opened 2024-06-12
Add overload for kind + identity in `useRemoteParticipant` hook

lukasIO:
Simplifies agent usecases

```ts
const agent = useRemoteParticipant({ kind: ParticipantKind.Agent });
```


davidzhao:
Thinking out loud, should we instead do this?

```
useParticipant(kind: ParticipantKind.Agent, name: "myagent", identity: "blah")
```

that way, we could provide a general framework of returning the first participant matching these key filters

lukasIO:
Today there's `useRemoteParticipant(identity)` already. 

I could imagine an overload for that hook with something like
```ts
useRemoteParticipant( { identity?: string; kind?: ParticipantKind })
```

What usecase do you imagine with `name` additionally being part of the selector/filter?

davidzhao:
> Today there's `useRemoteParticipant(identity)` already.
> 
> I could imagine an overload for that hook with something like
> 
> ```ts
> useRemoteParticipant( { identity?: string; kind?: ParticipantKind })
> ```
> 
> What usecase do you imagine with `name` additionally being part of the selector/filter?

great. I like that.. just thinking we could let user specify additional filters. Agreed name shouldn't be one today, I'm not quite sure why I put that in.

lukasIO:
Makes sense, updated the PR to add an overload for `useRemoteParticipant` instead

davidzhao (approved):
lg!

PR #917, opened 2024-07-09
Add useVoiceAssistant

lukasIO:
It needs more testing, but would be good to align on the general API first.


packages/react/etc/components-react.api.md
```diff
@@ -1175,6 +1185,24 @@ export interface VideoTrackProps extends React_2.VideoHTMLAttributes<HTMLVideoEl
     trackRef?: TrackReference;
 }
 
+// @alpha (undocumented)
+export interface VoiceAssistant {
+    // (undocumented)
+    agent: RemoteParticipant | undefined;
+    // (undocumented)
+    agentTranscriptions: ReceivedTranscriptionSegment[];
```
bcherry, line 1193: this would be the agent's text, right? the transcribed user audio would still be to be gotten out of a different place?

lukasIO: yeah, the user's transcriptions would still be tied to that user, via `useTrackTranscriptions(localParticipantAudioTrack)`

lukasIO: do you think it would make more sense to include local participant transcriptions + tracks here as well?

bcherry: not sure - would want to see how it works in practice. i think threading the agent and user transcriptions against eachother in 1:1 setting (and even potentially in non-1:1 settings) will be a common case but i don't think the way it's set up is a problem

packages/react/src/hooks/useVoiceAssistant.ts
```diff
@@ -0,0 +1,74 @@
...
+      return 'offline';
+    } else if (connectionState !== 'connected') {
+      return 'connecting';
+    }
+
+    if (!agent) {
+      return 'connecting';
+    }
+
+    if (!agentMetadata) {
```
davidzhao, line 51: can we use participant attributes for this? `agent.state`

packages/react/src/hooks/useVoiceAssistant.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * @alpha
+ *
+ * This hook looks for the first agent-participant in the room.
+ * It assumes that the agent participant is based on the LiveKit VoiceAssistant API and
+ * returns the most commonly used state vars when interacting with a VoiceAssistant.
+ */
+export function useVoiceAssistant(): VoiceAssistant {
+  const agent = useRemoteParticipants().find((p) => p.kind === ParticipantKind.AGENT);
+  const audioTrack = useParticipantTracks([Track.Source.Microphone], agent?.identity)[0];
```
davidzhao, line 35: if it hasn't been published yet, would this be accessing an invalid element of array?

lukasIO: audiotrack would be `undefined`, but not raising an exception

packages/react/src/hooks/useVoiceAssistant.ts
lukasIO: Agree this would be better.
I guess ultimately the approach for state has to be decided by the `VoiceAssistant` implementation within the agents framework server side.
Right now this client implementation is mirroring the behaviour of Neil's HP agent in order to have a working prototype.

packages/react/src/hooks/useVoiceAssistant.ts
davidzhao: good to know! some languages would raise.

packages/react/src/hooks/useVoiceAssistant.ts
davidzhao: I would advocate for changing the agent to use the new protocol that you are defining. would like to steer users into the right path for attributes vs parsing metadata

packages/react/etc/components-react.api.md
davidzhao: I like this idea.. I wonder if they should be combined at the protocol level? (i.e. renaming Transcriptions to be generic TextSegment? and sending chat through it as well?)

lukasIO: That's an interesting idea! 
I like that this would mean a protocol-level definition for chat messages (which we currently do not have). 
But I think it would still be good to keep transcriptions and chat messages distinguishable. 
Especially when thinking about a first class chat in the future.

bcherry: +1 to this concept - with some sort of type/kind/source flag that indicates where the text came from (direct input, transcribed from audio, read from OCR, etc)

And for what it's worth when building my own app on this sort of thing it was really important to distinguish between spoken-then-transcribed text and typed-text when using it in LLM inputs because you need to consider transcriptions to be of lower accuracy than typed inputs (e.g. if I typed "Lukas" in one spot and also spoke the name which got transcribed as "Lucas" somewhere else, the app needs to be able to know that the correct spelling is "Lukas").

packages/react/src/hooks/useVoiceAssistant.ts
lukasIO: got it, updated the PR to use attributes instead. 
This would require the agent side to set the state as `'listening' | 'thinking' | 'speaking'`. Not sure if we need anything else? (`connecting` and `disconnected` is inferred from the client's connection state and the agent's presence in the room)

packages/react/src/hooks/useTrackTranscription.ts
```diff
@@ -39,7 +39,7 @@ const TRACK_TRANSCRIPTION_DEFAULTS = {
  * @alpha
  */
 export function useTrackTranscription(
-  trackRef: TrackReferenceOrPlaceholder,
+  trackRef: TrackReferenceOrPlaceholder | undefined,
```
keepingitneil, line 42: 🙌

davidzhao (approved):
lgtm!

PR #931, opened 2024-07-26
Add useIsRecording hook

lukasIO:


davidzhao (approved):
hooks++

do we expose this to RN as well? where do we set that? CC: @davidliu

lukasIO:
[this](https://github.com/livekit/client-sdk-react-native/blob/main/src/hooks.ts) is were it's set for react-native. Would make sense to add it, once it's released here

PR #939, opened 2024-08-05
Add callback option to trackTranscription hook

lukasIO:
this simplifies usecases where users want to do their own state handling of incoming transcriptions (e.g. combining them in a timeline with chat messages)


bcherry (approved):

Ocupe (approved):
✍️

PR #940, opened 2024-08-05
Logarithmic binning of multiband track volumes

lukasIO:
for audio visualisation tasks it's more appropriate to use logarithmic binning of the linear frequency values in order to have a visual representation of frequency bands that more closely reflects human hearing.

### DIFF
--- a/packages/react/src/hooks/useTrackVolume.ts
+++ b/packages/react/src/hooks/useTrackVolume.ts
@@ -119,6 +119,7 @@ export function useMultibandTrackVolume(
     if (!track || !track?.mediaStream) {
       return;
     }
+
     const { analyser, cleanup } = createAudioAnalyser(track, opts.analyserOptions);
 
     const bufferLength = analyser.frequencyBinCount;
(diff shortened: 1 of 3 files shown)


packages/react/src/hooks/useTrackVolume.ts
```diff
@@ -154,3 +153,51 @@ export function useMultibandTrackVolume(
 
   return frequencyBands;
 }
+
+function calculateLogBinEdges(minFreq: number, maxFreq: number, numBins: number): number[] {
```
boks1971, line 157: probably worth looking at Bark Scale  also (https://en.wikipedia.org/wiki/Bark_scale) as it maps psychoacoustics.

lukasIO: oh, nice! might be even better!

boks1971: Mel Scale - https://en.wikipedia.org/wiki/Mel_scale may be better.

boks1971: One is pitch and the other is loudness scale.

Ocupe:
@lukasIO I guess this is just for nicer looking visuals, right? Do you maybe have a video showing the differences to the current implementation?

PR #943, opened 2024-08-06
chore: Update locale handling in ChatEntry component

Ocupe:
Currently all timestamps are displayed in the US format AM/PM.

This PR fixes that.
image


Ocupe:
Nevermind, seems to be an update bug with my Chrome test setup.

PR #946, opened 2024-08-15
Tweaks to voice assistant hook

lukasIO:


packages/react/src/hooks/useVoiceAssistant.ts
```diff
@@ -49,7 +50,7 @@ export function useVoiceAssistant(): VoiceAssistant {
     } else if (connectionState === ConnectionState.Connecting || !agent || !attributes?.state) {
       return 'connecting';
     } else {
-      return attributes.state as VoiceAssistantState;
+      return attributes['agent.state'] as VoiceAssistantState;
```
davidzhao, line 53: if this is not yet set. could we return a sane default instead of 'undefined'?

davidzhao (approved):

packages/react/src/hooks/useVoiceAssistant.ts
lukasIO: hm, wanted to just return `listening` but this causes a couple of minor headaches:

- the condition above would usually say `as long as agent.state has not been set, show the state as "connecting"`. I think this makes sense under the assumption that the first state the agent will set is `initializing`, so showing `connecting` before sounds reasonable.
- if we were to show `listening` as a default fallback, the agent might be going from `listening` to `initializing` which sounds rather unexpected

davidzhao: just thinking about this. if the server-side isn't setting this field, it would end up showing `connecting`, which seems a bit confusing if the agent has already connected. is `initializing` useful? i.e. why would we want the agent to be available in the room, but `initializing` instead of `connecting`?

lukasIO: > which seems a bit confusing if the agent has already connected

the expectation is that an agent would set `agent.state` attribute immediately upon connection. Would it still be a confusing for you in that case?

> is initializing useful?

Neil raised the point that an agent might still need to load/setup certain things, during which it is in fact connected, but is not yet in a state where it can `listen` to the user. 
Your suggestion seems to be to show `connecting` in that case, which you said above is confusing if the agent is in fact already connected.

lukasIO: FWIW, visually speaking it's still to be decided whether or not an AgentVisualizer would show `initializing` as a dedicated state, but it gives devs a bit more flexibility/insight into what's currently going on.

keepingitneil: So when discussing in slack I missed that there would a `connecting` state when I made my suggestion. I think `connecting` would be a sane default for the undefined. Basically it's the state that represents "not yet accepting user input".

It'd be up to users who have a custom agent to set this attribute themselves if they choose to use this hook client side.

lukasIO: after further discussion with @davidzhao and @bcherry we settled on keeping `initializing` as a state in the hook, but potentially dropping it for any visualisation of the voice assistant.

PR #949, opened 2024-08-28
Fix local storage of user choices

lukasIO:


packages/react/src/hooks/useTrackToggle.ts
```diff
@@ -65,7 +65,7 @@ export function useTrackToggle<T extends ToggleSource>({
   const clickHandler: React.MouseEventHandler<HTMLButtonElement> = React.useCallback(
     (evt) => {
       userInteractionRef.current = true;
-      toggle().finally(() => (userInteractionRef.current = false));
```
lukasIO, line 68: resetting this to `false` was happening before the `onChange` callback was fired in some instances. 
Removing the reset is fine, I think, because we reset it after every `onChange` call anyways.

If we want to make sure, we could reset it in `catch` instead of in `finally`

packages/react/src/hooks/useTrackToggle.ts
```diff
@@ -65,7 +65,7 @@ export function useTrackToggle<T extends ToggleSource>({
...
+      toggle().catch(() => (userInteractionRef.current = false));
```
lukasIO, line 68: we're only interested in resetting this on failures, as otherwise it will be reset above in the `useEffect`. 

We don't want to use `finally` here as it introduces a race around the emitted value of the `userInitiated` boolean

Ocupe (approved):
Nice!

PR #954, opened 2024-09-06
Stateful BarVisualizer and VoiceAssistantControlBar

lukasIO:
go to localhost:3000/voice-assistant for the demo

`<BarVisualizer />` can also be used _without_ passing voice assistant state and will then just behave like a regular bar visualizer without any state transitions (this is e.g. used for the bars in the control bar)


examples/nextjs/pages/agent.tsx
```diff
@@ -0,0 +1,77 @@
...
+        identity: userId,
+        name: userId,
+      },
+    };
+  }, []);
+
+  const token = useToken(process.env.NEXT_PUBLIC_LK_TOKEN_ENDPOINT, roomName, tokenOptions);
+
+  function AgentTile() {
+    const { state, audioTrack } = useVoiceAssistant();
```
bcherry, line 36: @lukasIO I'm assuming that we need to make AgentTile its own component because `useVoiceAssistant()` has to run inside of a `LiveKitRoom` context? If so - it seems a little clunky, no? Would it be possible for `AgentBarVisualizer` to just do this internally somehow? it would be nice if you could just do this:

```
<LiveKitRoom ...>
  <VoiceAssistantBarVisualizer />
  <VoiceAssistantControlBar />
</LiveKitRoom>
```

bcherry (approved):
this is great, thank you!

I got a little lost setting up the demo because my test agent app was on agents `0.8.5` so it didn't implement voice assistant state. but the demo app and component just went ahead anyways and it ended up permanently in the rapid left/right scanning state but also did the size transforms. 

I'm wondering if there's something we should be doing with setting an agent framework version as a participant attribute or something on the agent participant, and then maybe logging a warning or something? it was broken just enough to be a bad experience but not obviously enough for me to realize something was wrong. maybe not a common experience, though, most devs are new on both sides.

lukasIO:
That's a very good point. It might make sense to set an `voice-assistant-protocol` version. That way we're not tied to specific agent version numbers and only have to remember to bump it once something that actually impacts the agent to client behaviour wrt to voice agents?

davidzhao:
maybe the assistant could set some sort of capabilities? i.e. `assistant.capability.state = true` ?

davidzhao (commented):
nice work!

lukasIO:
setting a boolean would make it harder to track changes within the state protocol. In case we add/remove states in the future a version number would allow us to keep backwards compatibility

packages/styles/scss/components/participant/_audio-visualizer.scss
```diff
@@ -1,11 +1,80 @@
...
+  height: 100%;
+  background: var(--bg);
+  gap: var(--va-bar-gap, 24px);
+
+  & > .audio-bar {
+    transform-origin: 'center';
+    height: 100%;
+    width: var(--va-bar-width, 12px);
+    border-radius: var(--va-bar-border-radius, 32px);
+    background-color: rgba(255, 255, 255, 0.05);
```
bcherry, line 46: @lukasIO maybe i'm just missing something (still new to react + tailwind etc) but how do i customize the bar color? I want it in gray/black on white. Is the only option to write a css rule for `.lk-audio-bar { background-color }`?

lukasIO: ah, yeah, the un-highlighted bar colour is not exposed as a css variable right now. 
for hightlighted states you can use `--lk-fg`. Will add another one for the non-highlighted color state

PR #955, opened 2024-09-10
Add prepareConnection option to LiveKitRoom

lukasIO:
will call prepareConnection as soon as token + serverURL are set, independently of whether or not `connect` has been set to true on the room.


bcherry (approved):

1egoman:
Going to close this given the move away from recommending `LiveKitRoom`.

PR #976, opened 2024-09-17
Rename agent state property to lk.agent.state

bcherry:
Some back and forth here but we've landed on genericizing the state property and leaving the rest as-is.

### DIFF
--- a/.changeset/cuddly-beers-cry.md
+++ b/.changeset/cuddly-beers-cry.md
@@ -1,0 +1,6 @@
+---
+"@livekit/components-react": minor
+"@livekit/component-example-next": patch
+---
+
+Rename VoiceAssistantState to AgentState, VoiceAssistantControlBar to AssistantControlBar
--- a/.changeset/cuddly-beers-cry.md
+++ b/.changeset/cuddly-beers-cry.md
@@ -3,4 +3,4 @@
 "@livekit/component-example-next": patch
 ---
 
-Rename VoiceAssistantState to AgentState, VoiceAssistantControlBar to AssistantControlBar
+Update agent state property name to `lk.agent.state`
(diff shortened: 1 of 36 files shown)


packages/react/src/hooks/useAgent.ts
```diff
@@ -11,61 +11,60 @@ import { useParticipantAttributes } from './useParticipantAttributes';
...
-export type VoiceAssistantState =
+export type AgentState =
   | 'disconnected'
   | 'connecting'
   | 'initializing'
   | 'listening'
   | 'thinking'
-  | 'speaking';
+  | 'speaking'
+  | string;
```
bcherry, line 21: this change allows users to define their own agent states not included in our default enumeration and still take advantage of the `useAgent` hook

.changeset/cuddly-beers-cry.md
```diff
@@ -0,0 +1,6 @@
+---
+"@livekit/components-react": patch
```
lukasIO, line 2: even if in beta, my hunch is we should mark this as `minor`

```suggestion
"@livekit/components-react": minor
```

packages/react/src/hooks/useAgent.ts
bcherry: @lukasIO I changed this back too to keep it simple, but I'm wondering:

what happens if the backend sends a value not in this list, but the frontend expects only one of these five types? I'm assuming nothing too bad but I'm not really sure. We will definitely add more types in the future, in one way or another.

lukasIO (approved):

packages/react/src/hooks/useAgent.ts
lukasIO: that should be fine, others will just be ignored

PR #978, opened 2024-09-19
Expose StartMediaButton for External Use

junsumida:
Hello, LiveKit developers,

I noticed that the [`StartMediaButton`](https://github.com/livekit/components-js/blob/main/packages/react/src/components/controls/StartMediaButton.tsx#L23) component is marked as `@public`, but it’s currently not accessible from outside the package. It would be greatly appreciated if you could add it to the exported components list to make it available for external use. (I am currently customizing the control bar.)

Thank you for considering this request!

### DIFF
--- a/.changeset/modern-tips-eat.md
+++ b/.changeset/modern-tips-eat.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-react": patch
+---
+
+Add StartMediaButton to public API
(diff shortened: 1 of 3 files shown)


lukasIO:
thanks for catching this, yeah, we simply forgot to expose it! 
Happy to merge this once you signed the CLA

lukasIO (approved):

junsumida:
Thanks for your prompt response as always :smile:

PR #979, opened 2024-09-19
Add support for new chat API

lukasIO:
depends on https://github.com/livekit/client-sdk-js/pull/1224

### DIFF
--- a/packages/core/src/components/chat.ts
+++ b/packages/core/src/components/chat.ts
@@ -1,31 +1,49 @@
 /* eslint-disable camelcase */
-import type { Participant, Room } from 'livekit-client';
+import type { Participant, Room, ChatMessage } from 'livekit-client';
 import { RoomEvent } from 'livekit-client';
-import { BehaviorSubject, Subject, scan, map, takeUntil } from 'rxjs';
-import { DataTopic, sendMessage, setupDataMessageHandler } from '../observables/dataChannel';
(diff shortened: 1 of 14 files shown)


packages/core/src/components/chat.ts
```diff
@@ -154,5 +181,16 @@ export function setupChat(room: Room, options?: ChatOptions) {
...
-  return { messageObservable: messagesObservable, isSendingObservable: isSending$, send, update };
+  return {
+    messageObservable: messagesObservable,
+    isSendingObservable: isSending$,
+    send,
+    update,
+  };
+}
+
+function isIgnorableChatMessage(
```
bcherry, line 192: it might make more sense to filter these in the base SDK instead - or in addition - so devs aren't confused if they see both go by?

lukasIO: the base SDK does not have a concept of the legacy chat though. This was all defined within the components framework.

bcherry: if your project uses the new chat protocol, whether it uses components or not, won't it be sending and receiving both types? and the old one goes down the data channel as a user packet so if you listen to user packets on the data channel you'll see them and it might confuse you. Might be better for the base SDK to just drop them before they reach the user's own handlers?

lukasIO: > whether it uses components or not, won't it be sending and receiving both types?

it will only send both types if you're using components

PR #980, opened 2024-09-20
Update useParticipants to listen for any Participant's info changes

lebaudantoine:
### Issue
This is my first contribution! I’d love any advice on how to contribute to the repo. While working with the `useParticipants` hook, I expected it to keep the participants list updated with their names. However, I found that `useParticipants` only listens to metadata changes by default and does not react to updates in names or attributes.

To ensure the participants list updates correctly, I called the hook as follows:

```javascript
const participants = useParticipants({
  updateOnlyOn: [
    RoomEvent.ParticipantNameChanged,
    ...allParticipantRoomEvents,
  ],
});
```

Additionally, I noticed that the term `updateOnlyOn` could be misleading, as it always includes three events. While the documentation clarifies this, it may cause confusion when reading the code.

### Solution

I found the previous implementation a bit convoluted. I propose that `useParticipants` should update on all room events that trigger state changes for a participant, as originally intended by @Ocupe. Additionally, please note that a few events related to end-to-end encryption and permissions are still missing from the current setup.

### DIFF
--- a/.changeset/calm-rivers-behave.md
+++ b/.changeset/calm-rivers-behave.md
@@ -1,0 +1,5 @@
+---
+'@livekit/components-core': patch
+---
+
+update useParticipants to listen for any Participant's info changes
--- a/packages/core/src/helper/eventGroups.ts
+++ b/packages/core/src/helper/eventGroups.ts
@@ -11,6 +11,8 @@ export const allRemoteParticipantRoomEvents = [
   RoomEvent.ParticipantDisconnected,
   RoomEvent.ParticipantPermissionsChanged,
   RoomEvent.ParticipantMetadataChanged,
+  RoomEvent.ParticipantNameChanged,
+  RoomEvent.ParticipantAttributesChanged,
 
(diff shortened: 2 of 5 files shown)


lebaudantoine:
what do you think @lukasIO?

lukasIO (approved):

.changeset/calm-rivers-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@livekit/components-core': minor
```
lukasIO, line 2:
```suggestion
'@livekit/components-core': patch
```

as it's primarily a bugfix, I think we should mark it as a patch update

lukasIO:
@lebaudantoine feel free to open a subsequent PRs for the other missing events, we'd definitely want them to be part of `allRemoteParticipantEvents` as well!

.changeset/calm-rivers-behave.md
lebaudantoine: Fixed, ready for merge!

PR #983, opened 2024-09-24
fix jsx tags in example comments

paulwe:
https://github.com/livekit/livekit/issues/3000


davidzhao (approved):
wow.. thanks for the fix!

PR #985, opened 2024-09-25
Add css var for idle visualizer bar color

lukasIO:


bcherry (approved):
Consider changing the default colors to be a faint gray rather than a translucent white, so it's not invisible on a white bg. right now the experience of dropping this into a project with a white bg is pretty confusing, since you need to open the inspector to see if it even worked.  cc @mattherzog

PR #986, opened 2024-09-27
Add krisp hook

lukasIO:


bcherry (approved):

davidzhao (approved):
awesome. let's get this documented afterwards too, would be much easier than the current path

PR #987, opened 2024-09-29
Improves bar visualizer animations and adds initializing state

mattherzog:
```
['connecting', 2000],
['initializing', 2000],
```

- Updated connecting interval to smooth out animation. These intervals shouldn't really be intervals as they are divided by `barCount`.  What we actually want is to define the overall length of the animation, in this case 2s, and divide that by `barCount`. This code works, but feels a bit indirect to call it 'intervals' with this update.

- Added an entry for 'initializing' as this wasn't supported and produced animation glitching. 

- Updated the connection sequence as to not repeat first/last when animation repeats.

### DIFF
--- a/.changeset/ninety-cheetahs-tickle.md
+++ b/.changeset/ninety-cheetahs-tickle.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-react": patch
+---
+
+Improve bar visualizer animations and add initializing state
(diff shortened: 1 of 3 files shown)


mattherzog:
https://github.com/user-attachments/assets/a9b7ca19-4e95-408c-a080-2a3387cc2966

lukasIO (approved):
🧈

davidzhao (approved):

PR #1014, opened 2024-10-22
Fix initial device selection when mounting useMediaDeviceSelect

lukasIO:


packages/core/src/observables/room.ts
```diff
@@ -249,7 +249,6 @@ export function createActiveDeviceObservable(room: Room, kind: MediaDeviceKind)
       log.debug('activeDeviceObservable | RoomEvent.ActiveDeviceChanged', { kind, deviceId });
       return deviceId;
     }),
-    startWith(room.getActiveDevice(kind)),
```
lukasIO, line 252: this causes the active device to change whenever the `useMediaDeviceSelect` hook is mounted. Instead, shift the responsibility for initializing with the correct value into the hook's initial `currentDeviceId` state, which has been done already in https://github.com/livekit/components-js/pull/1012/files#diff-246daab5d71d5f9448a651e07a59775cbdccae8b7863e9f289fab063686772d1R57

Ocupe (approved):
👍

PR #1029, opened 2024-11-20
React requires the class property to be called 'className'

mikljohansson:
Reproduce errors as 

```
<BarVisualizer
  trackRef={{participant, publication, source: publication?.source || Track.Source.Microphone}}
  barCount={3}>
  <span>
    <span style={{backgroundColor: "black"}}/>
  </span>
</BarVisualizer>
```

Adding `<span>'s` as children like this is required to override the theming for this component


lukasIO (approved):
thanks for catching that!

PR #1038, opened 2024-12-02
Improve doc gen

Ocupe:
- Inline deprecation warning to improve rendering.

### DIFF
--- a/tooling/api-documenter/src/documenters/MarkdownDocumenter.ts
+++ b/tooling/api-documenter/src/documenters/MarkdownDocumenter.ts
@@ -225,13 +225,38 @@ export class MarkdownDocumenter {
         );
 
         if (tsdocComment.deprecatedBlock) {
+          // Render the deprecation message as a "caution" callout add a "deprecated" prefix to the message.
+          let addedDeprecationString = false;
+          const newContent = tsdocComment.deprecatedBlock.content.nodes.reduce<DocNode[]>(
+            (acc, node) => {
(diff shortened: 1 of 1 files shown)


tooling/api-documenter/src/documenters/MarkdownDocumenter.ts
```diff
@@ -225,13 +225,38 @@ export class MarkdownDocumenter {
...
 
         if (tsdocComment.deprecatedBlock) {
+          // Render the deprecation message as a "caution" callout add a "deprecated" prefix to the message.
+          let addedDeprecationString = false;
+          const newContent = tsdocComment.deprecatedBlock.content.nodes.reduce<DocNode[]>(
+            (acc, node) => {
+              if (!addedDeprecationString && node.kind === DocNodeKind.Paragraph) {
+                const paragraphChildren = node
+                  .getChildNodes()
+                  .reduce<DocNode[]>((paragraphNodes, paraNode) => {
```
lukasIO, line 235: is this omitting some elements? `reduce` vs `map` always makes me expect a non equivalent number of items being returned, same question above in L230

Ocupe: It is expected to return the same number of elements. I used reduce to iterate over the nodes and once we update the first Plain Text node we skip the first code branch and just push the elements onto the array.

lukasIO (approved):

PR #1042, opened 2024-12-06
Fix options stringify crash

toger5:
We cannot stringify options if it has a track processor. The processor is to large and we will end up exceeding string size limits.


toger5:
Close in favor of: https://github.com/livekit/components-js/pull/1039

PR #1043, opened 2024-12-06
use options instead of strigified options

toger5:
This is another option to fix the issue, that stringify does not work on the large (optional) options.video.processor object.

Here we just only ever update if the options themselves change (no deep comp).
This could be fine but makes the requirement, that the user known they have to construct new options if they want to update the tracks.


toger5:
Closed in favor of https://github.com/livekit/components-js/pull/1039

PR #1044, opened 2024-12-07
Use shared logger for krisp hook

bcherry:
This ensures it respects your log levels set on the SDK/components. I also clarified a message in case it helps for people who don't know what any of this stuff is 🤷🏻


lukasIO (approved):
nice!

bcherry:
not sure why changeset bot didn't update the status but i added a changeset a few days ago

PR #1050, opened 2024-12-14
use JS SDK's isLocal check

### LINKED ISSUE #1037: RoomAudioRenderer also rendered local track

\@livekit/components-react

### Describe the bug

I found the filter in RoomAudioRenderer is not filter the local track, I add a console log for tracks:
```tsx
  const tracks = useTracks(
    [Track.Source.Microphone, Track.Source.ScreenShareAudio, Track.Source.Unknown],
    {
      updateOnlyOn: [],
      onlySubscribed: true,
    },
  ).filter((ref) => !isLocal(ref.participant) && ref.publication.kind === Track.Kind.Audio);

  // my log
  console.warn(
    tracks,
    tracks[0]?.participant,
    tracks[0]?.participant.isLocal,
    isLocal(tracks[0]?.participant),
  );
```
and got follow output:
image

the value of `participant.isLocal` and `isLocal(participant)` is different.
But I found the meet.livekit.io web demo do not have this issue, and I use livekit in electron enviroment. It should be a enviroment compatibility problem ?

### Reproduction

use livekit components in electron app

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 15.2
    CPU: (12) arm64 Apple M3 Pro
    Memory: 204.58 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.18.0 - ~/.nvm/versions/node/v20.18.0/bin/node
    Yarn: 1.22.22 - ~/Desktop/chative-desktop/node_modules/.bin/yarn
    npm: 10.8.2 - ~/.nvm/versions/node/v20.18.0/bin/npm
  Browsers:
    Chrome: 131.0.6778.86
    Firefox: 130.0
    Safari: 18.2
  npmPackages:
    livekit-client: ^2.5.10 => 2.6.3
```

### Severity

serious, but I can work around it

### Additional Information

_No response_

davidzhao:
closes #1037

while it's not totally clear why instance type check failed in this example. it seems like a good practice to us the JS SDK's isLocal check.

### DIFF
--- a/.changeset/wise-ghosts-talk.md
+++ b/.changeset/wise-ghosts-talk.md
@@ -1,0 +1,6 @@
+---
+"@livekit/components-core": patch
+"@livekit/components-react": patch
+---
+
+use JS SDK's isLocal check
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -1,16 +1,15 @@
 import type { Participant, TrackPublication } from 'livekit-client';
-import { LocalParticipant, RemoteParticipant } from 'livekit-client';
 
-import type { PinState } from './types';
 import type { TrackReference } from './track-reference';
 import { isEqualTrackRef } from './track-reference';
+import type { PinState } from './types';
 
 export function isLocal(p: Participant) {
-  return p instanceof LocalParticipant;
+  return p.isLocal;
 }
 
 export function isRemote(p: Participant) {
-  return p instanceof RemoteParticipant;
+  return !p.isLocal;
 }
 
 export const attachIfSubscribed = (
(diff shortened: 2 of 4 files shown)


lukasIO (approved):
oh, that's great!

the fact that `instance of` is failing does sound like an environment issue, where multiple versions of the `livekit-client` package are being used (e.g. cjs and esm). That's indicative of the dual package hazard.

davidzhao:
that makes sense.. I remember seeing this in React-land when overriding for local test.. it was confusing and I've never figured it out..

PR #1051, opened 2024-12-14
Replace instanceof LocalParticipant checks

lukasIO:
follow up to #1050. 

There are still additional places where we perform `instanceof` checks for tracks and track publications. To remove those we'll first need an update to the client sdk exposing something like `participant.isLocal` on tracks and publications.


davidzhao (approved):
nice. good call on checking this

PR #1052, opened 2024-12-17
preserve casing in generated docs page slugs

bcherry:
right now these get lowercased which makes the links less readable

### DIFF
--- a/.changeset/wicked-masks-serve.md
+++ b/.changeset/wicked-masks-serve.md
@@ -1,0 +1,5 @@
+---
+"@livekit/api-documenter": patch
+---
+
+preserve casing in generated docs page slugs
--- a/tooling/api-documenter/src/utils/Utilities.ts
+++ b/tooling/api-documenter/src/utils/Utilities.ts
@@ -21,6 +21,6 @@ export class Utilities {
   public static getSafeFilenameForName(name: string): string {
     // TODO: This can introduce naming collisions.
     // We will fix that as part of https://github.com/microsoft/rushstack/issues/1308
-    return name.replace(Utilities._badFilenameCharsRegExp, '_').toLowerCase();
+    return name.replace(Utilities._badFilenameCharsRegExp, '_');
   }
 }


Ocupe (commented):
I can understand the motivation, but I don't think the gain is worth the cost (on the docs site) of the change. 
This is of course a discussion about tased, but I don't particularly like case sensitive urls and would rather change it to kebab-case if we have to change it at all.

Ocupe:
> @Ocupe next.js URLs are already case sensitive (see https://docs.livekit.io/reference/components/react/component/audiotrack/ which works, but https://docs.livekit.io/reference/components/react/component/AUDIOTRACK/ is a 404), this just applies a more appropriate casing.

@bcherry  I see my mistake now. I meant to say that I don't like the look of urls with non-uppercase characters. But as I said, that is just my taste.

> From what I see, most reference docs maintain casing exactly as is in the original language:
> 
> React's own docs keep original casing: https://react.dev/reference/react/useDeferredValue Python generated docs also keep casing, but its more of a mess anyways: https://docs.livekit.io/python/livekit/agents/stt/fallback_adapter.html#livekit.agents.stt.fallback_adapter.AvailabilityChangedEvent and typedoc keeps casing too https://docs.livekit.io/client-sdk-js/enums/KeyProviderEvent.html
> 
> two exceptions I found are swift/apple, which are all lowercase: https://developer.apple.com/documentation/avfoundation/avplayer/init(playeritem:)
> 
> and kotlin/android, which are indeed kebab-case https://docs.livekit.io/components-android/livekit-compose-components/io.livekit.android.compose.state.transcriptions/remember-participant-transcriptions.html
> 
> I'm fine with going to kebab case here if that's what people prefer, but I don't love the current all lowercase...


Thanks for listing some examples from other docs, I appreciate the overview!
My feeling is that most people don't really care about this, so why change it when we have the burden of backwards compatibility on the docs site?
That being said, and because I don't have a strong opinion (just a slight dislike :P), I'll leave the decision up to you.

Ocupe (approved):

bcherry:
@Ocupe i think you're right, i'll switch to kebab case for react. that's what russ asked for in slack originally but i wasn't convinced until cursor autocompleted to expect urls in kebab case in a markdown doc i was writing... :)

PR #1055, opened 2024-12-18
Include useKrispNoiseFilter in generated docs, + related improvements

bcherry:
The main issue here is the krisp hook was excluded from generated docs because it's not actually in the react package index.d.ts. I decided the most straightforward thing to do is include it in generated docs anyways alongside everything else. I improved its docs a bit and upgraded it from alpha to beta.

While working on this I found and fixed the following additional issues:

- Added a custom `@package` tag for tsdoc so we can override the import path for the krisp hook
- Turn the `@beta` tag into a callout instead of a paragraph (and some text edits)
- Add a callout for the `@alpha` tag just like the `@beta` tag has
- We automatically unrolled parameters with type names ending in "Props" but we also sometimes use "Options" instead, including for useKrispNoiseFilter, so I added that.
- Unrolled parameters are confusing when there are other parameters in the list as well so I also added a prefix to them indicating where they should be included (e.g. `options.abc`). this still has room for improvement
- The `@return` directive is currently ignored entirely in favor of a generated signature. To improve docs, I've added the return description, if supplied, alongside the generated code.
- Multi-example docs were kinda unnatural with totally separate sections labeled "Usage Example 1", "Usage Example 2", etc. I changed it to be one section called Usage, with sub-headers for individual examples. They now use their first line as the title, falling back to numbered examples. I added extra examples to useIsMuted with this change.

image

image

### DIFF
--- a/packages/react/src/hooks/cloud/krisp/useKrispNoiseFilter.ts
+++ b/packages/react/src/hooks/cloud/krisp/useKrispNoiseFilter.ts
@@ -10,18 +10,22 @@ import { useLocalParticipant } from '../../..';
  */
 export interface useKrispNoiseFilterOptions {
   /**
-   * by default the hook will use the localParticipant's microphone track publication.
-   * You can override this behavior by passing in a target TrackReference here
+   * The track reference to use for the noise filter (defaults: local microphone track)
    */
   trackRef?: TrackReferenceOrPlaceholder;
+  /**
+   * @internal
+   */
   filterOptions?: NoiseFilterOptions;
 }
 
 /**
  * Enable the Krisp enhanced noise cancellation feature for local audio tracks.
  *
+ * Defaults to the localParticipant's microphone track publication, but you can override this behavior by passing in a different track reference.
+ *
  * @package @livekit/components-react/krisp
- * @remarks This filter requires that you installthe `@livekit/krisp-noise-filter` package and is supported only on [LiveKit Cloud](https://cloud.livekit.io).
+ * @remarks This filter requires that you install the `@livekit/krisp-noise-filter` package and is supported only on [LiveKit Cloud](https://cloud.livekit.io).
  * @beta
  * @example
  * ```tsx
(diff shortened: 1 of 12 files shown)


packages/react/src/hooks/cloud/krisp/useKrispNoiseFilter.ts
```diff
@@ -6,33 +6,38 @@ import type { TrackReferenceOrPlaceholder } from '@livekit/components-core';
...
  */
 export interface useKrispNoiseFilterOptions {
   /**
-   * by default the hook will use the localParticipant's microphone track publication.
-   * You can override this behavior by passing in a target TrackReference here
+   * The track reference to use for the noise filter (defaults: local microphone track)
    */
   trackRef?: TrackReferenceOrPlaceholder;
+  /**
+   * @internal
```
lukasIO, line 17: why would they be internal?

tooling/api-documenter/src/documenters/MarkdownDocumenter.ts
```diff
@@ -209,6 +208,9 @@ export class MarkdownDocumenter {
       if (apiItem.releaseTag === ReleaseTag.Beta) {
         this._writeBetaWarning(output);
       }
+      if (apiItem.releaseTag === ReleaseTag.Alpha) {
+        this._writeAlphaWarning(output);
```
lukasIO, line 212: nice!

lukasIO (approved):
looks good to me, but will defer to @Ocupe

packages/react/src/hooks/cloud/krisp/useKrispNoiseFilter.ts
bcherry: this was the easiest way to just leave it out of the docs. this comes from the closed-source underlying krisp SDK, so there's no good way to document this parameter. the actual values available to tweak here also seem unlikely to be useful, it's just debugging flags basically? I don't feel strongly about it I just thought it's easier to exclude it from the docs.

Ocupe (approved):
Looking good 👍

PR #1056, opened 2024-12-19
Improve default device handling

lukasIO:
components follow up for https://github.com/livekit/client-sdk-js/pull/1357


Ocupe (approved):
👍

PR #1069, opened 2025-01-21
Fix toggle hydration error

lukasIO:
track toggle would be server rendered and not updated some times leading to wrong UI state. 

This PR ensures the TrackToggle is only rendered on client side as recommended by react docs 🫠 
 https://nextjs.org/docs/messages/react-hydration-error#solution-1-using-useeffect-to-run-on-the-client-only

### DIFF
--- a/packages/react/src/prefabs/PreJoin.tsx
+++ b/packages/react/src/prefabs/PreJoin.tsx
@@ -232,12 +232,12 @@ export function PreJoin({
   ...htmlProps
 }: PreJoinProps) {
   // TODO: Remove and pipe `defaults` object directly into `usePersistentUserChoices` once we fully switch from type `LocalUserChoices` to `UserChoices`.
-  const partialDefaults: Partial<LocalUserChoices> = {
-    ...(defaults.audioDeviceId !== undefined && { audioDeviceId: defaults.audioDeviceId }),
-    ...(defaults.videoDeviceId !== undefined && { videoDeviceId: defaults.videoDeviceId }),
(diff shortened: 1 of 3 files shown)


packages/react/src/prefabs/PreJoin.tsx
```diff
@@ -231,17 +230,6 @@ export function PreJoin({
   videoProcessor,
   ...htmlProps
 }: PreJoinProps) {
-  const [userChoices, setUserChoices] = React.useState(defaultUserChoices);
-
-  // TODO: Remove and pipe `defaults` object directly into `usePersistentUserChoices` once we fully switch from type `LocalUserChoices` to `UserChoices`.
```
lukasIO, line 236: @Ocupe do you remember why this was handled in this particular way? 
Type wise nothing is complaining when simply piping the defaults directly into `usePersistentUserChoices`

Ocupe: I think the TODO note says why, right? We had this transition period with different/incompatible types.

Ocupe: I think because of the default/fallback parameter for `defaults` (line 220) it will trigger a constant rerender. Maybe it was a way to stop that from happening.

Ocupe (approved):

packages/react/src/prefabs/PreJoin.tsx
lukasIO: but both places are using the same type `LocalUserChoices` 🤔 
not sure where `UserChoices` is defined at all. Maybe this was an oversight when `UserChoices` got removed?

PR #1080, opened 2025-02-07
Fix useDataChannel's send return type

murilo-teleo:
Adjusting `useDataChannel` TypeScript definition to match the `send` function returned by 'setupDataMessageHandler`.


lukasIO (approved):
Thank you!

PR #1091, opened 2025-02-14
improve chat docs

bcherry:
I'm getting rid of https://docs.livekit.io/recipes/react-chat for now, although something like it may come back. So the content is being dispersed to here and the text streams doc.

### DIFF
--- a/packages/react/src/hooks/useChat.ts
+++ b/packages/react/src/hooks/useChat.ts
@@ -16,7 +16,6 @@ import { useConnectionState } from './useConnectionStatus';
  * @returns An object containing:
  * - `chatMessages` - Array of received chat messages
  * - `send` - Function to send a new message
- * - `update` - Function to edit an existing message
  * - `isSending` - Boolean indicating if a message is currently being sent
  *
  * @example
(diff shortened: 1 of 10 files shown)


packages/react/src/hooks/useChat.ts
```diff
@@ -8,8 +8,36 @@ import { useConnectionState } from './useConnectionStatus';
...
- * It also returns a `update` function that allows you to implement message-edit functionality.
+ * 
+ * @remarks
+ * Message history is not persisted and will be lost if the component is refreshed.
+ * You may want to persist message history in the browser, a cache or a database.
+ * 
+ * @returns An object containing:
+ * - `chatMessages` - Array of received chat messages
+ * - `send` - Function to send a new message
+ * - `update` - Function to edit an existing message
```
lukasIO, line 19: please leave this one out, we don't want to support it

lukasIO (approved):

lukasIO (approved):

lukasIO (approved):
eslint is still complaining, but lgtm otherwise

PR #1096, opened 2025-02-27
Add support for datastream based chat

lukasIO:


packages/react/src/hooks/useReceiveTextStream.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { TextStreamHandler } from 'livekit-client';
+
+/**
+ * @beta
+ * set up a text stream handler for the given topic
+ * @param topic - the topic to listen to
+ * @param onTextStreamReceived - the handler to call when a text stream is received
+ * @returns a function to unregister the text stream handler
+ */
+export function useReceiveTextStream(topic: string, onTextStreamReceived: TextStreamHandler) {
```
bcherry, line 12: maybe these methods would be better called `useHandleTextStream` to more closely align in name with the base feature

packages/react/src/hooks/useReceiveByteStream.ts
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import { useRoomContext } from '../context';
+import type { ByteStreamHandler } from 'livekit-client';
+
+/**
+ * @beta
+ * set up a byte stream handler for the given topic
```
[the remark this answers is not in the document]
bcherry: also include a note that it will throw an error if there's already a handler registered (you can only do this in one place at a time)

bcherry (approved):

packages/react/src/hooks/useReceiveByteStream.ts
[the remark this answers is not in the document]
lukasIO: removed this hook for now

packages/react/src/hooks/useReceiveTextStream.ts
lukasIO: it's called `useTextStream` now and automatically handles calls from multiple places, so it won't throw

packages/core/src/components/chat.ts
```diff
@@ -41,83 +41,87 @@ export type ChatOptions = {
...
-    const { messageObservable } = setupDataMessageHandler(room, [topic, updateTopic]);
-    messageObservable.pipe(takeUntil(onDestroyObservable)).subscribe(messageSubject);
+    room.registerTextStreamHandler(topic, async (reader, participantInfo) => {
+      const { id, timestamp } = reader.info;
+      const streamObservable = from(reader).pipe(
+        scan((acc: string, chunk: string) => {
+          return acc + chunk;
+        }),
+        map((chunk: string) => {
+          console.log('text stream updated', { id, chunk, attributes: reader.info.attributes });
```
bcherry, line 89: should these console.logs be removed or downgraded to debug?

bcherry (approved):

packages/core/src/components/chat.ts
lukasIO: thanks for catching that, removed

PR #1109, opened 2025-03-27
Add useTranscription hook

lukasIO:
with filtering based on participant identites and track ids.

### DIFF
--- a/packages/core/src/components/chat.ts
+++ b/packages/core/src/components/chat.ts
@@ -8,6 +8,7 @@ import {
   sendMessage,
   setupDataMessageHandler,
 } from '../observables/dataChannel';
+import { log } from '../logger';
 
 /** @public */
 export type { ChatMessage };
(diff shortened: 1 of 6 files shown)


packages/react/src/hooks/useTranscriptions.ts
```diff
@@ -0,0 +1,44 @@
...
+ * useTranscriptions is a hook that returns the transcriptions for the given participant identities and track sids,
+ * if no options are provided, it will return all transcriptions
+ * @example
+ * ```tsx
+ * const transcriptions = useTranscriptions();
+ * return <div>{transcriptions.map((transcription) => transcription.text)}</div>;
+ * ```
+ */
+export function useTranscriptions(opts?: UseTranscriptionsOptions) {
+  const { participantIdentities, trackSids } = opts ?? {};
```
bcherry, line 24: doesn't adding these options expose an easy trap where you have

```
const agentTranscriptions = useTranscriptions({ participantIdentities: ['agent'] });
```

and

```
const userTranscriptions = useTranscriptions({ participantIdentities: ['user'] });
```

in two different locations and then they conflict on `useTextStream` which can only be used once?

bcherry: ah ok that works. i do still think we should just make it easy for anyone to register multiple handlers but it's not a blocker to this PR

bcherry (approved):

bcherry (commented):
should this also get adopted by useVoiceAssistant?

PR #1117, opened 2025-04-07
Improve device menu auto placement

lukasIO:


packages/react/src/prefabs/MediaDeviceMenu.tsx
```diff
@@ -93,12 +96,12 @@ export function MediaDeviceMenu({
 
   React.useEffect(() => {
     document.addEventListener<'click'>('click', handleClickOutside);
-    window.addEventListener<'resize'>('resize', () => setUpdateRequired(true));
+    // window.addEventListener<'resize'>('resize', () => setUpdateRequired(true));
```
Ocupe, line 99: Is this intentional or should it be removed?

packages/react/src/prefabs/MediaDeviceMenu.tsx
```diff
@@ -93,12 +96,12 @@ export function MediaDeviceMenu({
...
     return () => {
       document.removeEventListener<'click'>('click', handleClickOutside);
-      window.removeEventListener<'resize'>('resize', () => setUpdateRequired(true));
+      // window.removeEventListener<'resize'>('resize', () => setUpdateRequired(true));
```
Ocupe, line 102: same

Ocupe (approved):
Not sure if we should remove the resize event listener.

packages/react/src/prefabs/MediaDeviceMenu.tsx
lukasIO: the idea is that the responsibility for this is being shifted to the `autoUpdate` helper of the floating ui lib. 
this includes screen resizing and a bunch of other things, see https://floating-ui.com/docs/autoupdate

PR #1129, opened 2025-04-15
Improve PreJoin Component and add custom participant placeholders

Simba14:
This PR improves the `PreJoin` component by fixing a few bugs, improving the UX and adding some new features.

Bug fixes:
- Video feed no longer flickers when enabling / disabling the audio (Due to Video stream being re-rendered each time)

Feature:
- Display selected devices
- Enabled selecting of device when audio / video is disabled. 
- Improve UI of selecting device section to make it clearer which part disables / enables and which part selects the device.
- Also extended the `ParticipantTile` by adding the placeholders prop. This allows placeholders to be unique for each participant e.g. if they have an image associated with their account. 

`placeholders` is an object where the key is the participant id and the value is a React node. 

In order to implement the selected device feature, I added new hook - `useSelectedDevice` to return the current selected device by the user.


packages/react/src/hooks/useSelectedDevice.ts
```diff
@@ -0,0 +1,72 @@
...
+ *   track: track,
+ * });
+ *
+ * <div>
+ *  {selectedDevice?.label}
+ * </div>
+ * ```
+ * @public
+ */
+export function useSelectedDevice<T extends LocalVideoTrack | LocalAudioTrack>({
```
lukasIO, line 22: within the livekit-client SDK there's the concept of an active device which is currently used by the user. 
That's not as useful for prejoin as of now, but it would mirror the implementation in the client SDK more closely.

packages/react/src/prefabs/PreJoin.tsx
```diff
@@ -59,46 +60,117 @@ export function usePreviewTracks(
...
 
-    return () => {
-      needsCleanup = true;
-      localTracks.forEach((track) => {
-        track.stop();
-      });
-    };
-  }, [JSON.stringify(options, roomOptionsStringifyReplacer), onError, trackLock]);
+  // Handle audio track
+  React.useEffect(() => {
```
lukasIO, line 155: the way it was previously implemented was in order to avoid separate permission prompts when entering the site. 
With your proposed implementation it would result in two different permission requests. 

I don't think that's an improvement worth the additional complexity.

packages/react/src/prefabs/PreJoin.tsx
```diff
@@ -362,41 +448,43 @@ export function PreJoin({
...
-              initialSelection={audioDeviceId}
-              kind="audioinput"
-              disabled={!audioTrack}
-              tracks={{ audioinput: audioTrack }}
-              onActiveDeviceChange={(_, id) => setAudioDeviceId(id)}
-            />
+          />
+          <div className="lk-button-group-menu-pre-join">
+            <label className="lk-selected-device-label">
+              {selectedAudioDevice?.label || micLabel}
```
lukasIO, line 459: this is a nice idea, but the handling could be moved into the <MediaDeviceMenu /> instead in order to allow for the same experience everywhere (not only in the prejoin). 
Keeping it simple with a `useDeviceInfo(track)` hook that only returns the MediaDeviceInfo for that use case seems more appropriate

packages/styles/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@livekit/components-styles",
-  "version": "1.1.4",
+  "version": "1.1.5",
```
lukasIO, line 3: versions are automatically bumped by the CI, bumping them in PRs has the effect of skipping versions or unintentionally publishing them

packages/styles/scss/prefabs/prejoin.scss
```diff
@@ -59,25 +60,20 @@
   }
 
   .button-group-container {
-    display: flex;
-    flex-wrap: nowrap;
+    display: grid;
+    grid-template-columns: repeat(2, minmax(0, 1fr));
```
lukasIO, line 64: these style changes aren't clear to me, why are they needed?

lukasIO (commented):
hi @Simba14 

thanks for taking the time to suggest and implement these changes. 
It's clear that you've put a lot of thought into it. 

I have some reservations about some of these changes though. 
I tried to outline my concerns in some comments, feel free to correct my assumptions if any sound wrong to you. 

The PR itself is (very) big and touches upon a bunch of things at the same time. We can keep this PR open to discuss your proposed changes. Whatever changes remain it would be nice to have them separated in PRs with smaller scope.

packages/react/src/components/participant/ParticipantTile.tsx
```diff
@@ -73,6 +73,7 @@ export interface ParticipantTileProps extends React.HTMLAttributes<HTMLDivElemen
   disableSpeakingIndicator?: boolean;
 
   onParticipantClick?: (event: ParticipantClickEvent) => void;
+  placeholders?: { [index: string]: React.ReactNode };
```
lukasIO, line 76: not a big fan of passing the whole map to each participant tile. 
A pattern where the component can take a single placeholder as a ReactNode seems preferable to me. 
In general the codebase tries to avoid passing of components as props and rather works with slottable components (as children). We're already using those for ParticipantTile as custom tile renderings though, so I'm not totally against an exception here.

PR #1130, opened 2025-04-22
Pull audio track from avatar worker in useVoiceAssistant

bcherry:
1. ensure avatar worker is never returned by useVoiceAssistant
2. check avatar worker for audio track if main agent doesn't have one
3. add video track to useVoiceAssistant

There's a remaining bug which is that useTrackTranscriptions doesn't work, as the agent won't publish legacy transcriptions if it has no audio track id. so we either need to fix something in the agents sdk so it can publish them anyways or just port this hook over to text streams

but for now this fixes the agent playground for video + audio at least

### DIFF
--- a/packages/react/src/hooks/useVoiceAssistant.ts
+++ b/packages/react/src/hooks/useVoiceAssistant.ts
@@ -26,6 +26,7 @@ export interface VoiceAssistant {
   agent: RemoteParticipant | undefined;
   state: AgentState;
   audioTrack: TrackReference | undefined;
+  videoTrack: TrackReference | undefined;
   agentTranscriptions: ReceivedTranscriptionSegment[];
   agentAttributes: RemoteParticipant['attributes'] | undefined;
 }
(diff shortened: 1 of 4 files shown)


packages/react/src/hooks/useVoiceAssistant.ts
```diff
@@ -42,8 +43,19 @@ const state_attribute = 'lk.agent.state';
  * @beta
  */
 export function useVoiceAssistant(): VoiceAssistant {
-  const agent = useRemoteParticipants().find((p) => p.kind === ParticipantKind.AGENT);
-  const audioTrack = useParticipantTracks([Track.Source.Microphone], agent?.identity)[0];
+  const agent = useRemoteParticipants().find(
+    (p) => p.kind === ParticipantKind.AGENT && !('lk.publish_on_behalf' in p.attributes),
+  );
+  const worker = useRemoteParticipants().find(
```
lukasIO, line 49: nit: no need to call `useRemoteParticipants` twice. Storing it once looks a lot cleaner. 
The hook sets up a bunch of internal event listeners, so it's nicer if we can avoid multiple unnecessary calls to it.

packages/react/src/hooks/useVoiceAssistant.ts
```diff
@@ -42,8 +43,19 @@ const state_attribute = 'lk.agent.state';
...
+    (p) =>
+      p.kind === ParticipantKind.AGENT && p.attributes['lk.publish_on_behalf'] === agent?.identity,
+  );
+  const agentAudioTrack = useParticipantTracks([Track.Source.Microphone], agent?.identity)[0];
```
lukasIO, line 53: similar notion as above, it's generally more efficient to filter on `useParticipantTracks(['microphone', 'camera'])` than calling it multiple times.

packages/react/src/hooks/useVoiceAssistant.ts
```diff
@@ -66,6 +78,7 @@ export function useVoiceAssistant(): VoiceAssistant {
     agent,
     state,
     audioTrack,
+    videoTrack,
```
lukasIO, line 81: wouldn't we simply want to return all tracks that the agent publishes? 
not a fan of `audioTrack` in the first place, but adding `videoTrack` doesn't make it any better. 
Generally we have `microphoneTrack` and `cameraTrack` in our glossary of constructs, but not sure how well that works for agents

lukasIO (approved):
just some nits and comments

packages/react/src/hooks/useVoiceAssistant.ts
```diff
@@ -42,8 +43,19 @@ const state_attribute = 'lk.agent.state';
  * @beta
  */
 export function useVoiceAssistant(): VoiceAssistant {
-  const agent = useRemoteParticipants().find((p) => p.kind === ParticipantKind.AGENT);
-  const audioTrack = useParticipantTracks([Track.Source.Microphone], agent?.identity)[0];
+  const agent = useRemoteParticipants().find(
+    (p) => p.kind === ParticipantKind.AGENT && !('lk.publish_on_behalf' in p.attributes),
```
lukasIO, line 47: where are we documenting these attributes?

packages/react/src/hooks/useVoiceAssistant.ts
bcherry: yeah i don't have a strong opinion really. this seemed least disruptive and i think if you really want all the tracks you can get them from `agent` anyways

packages/react/src/hooks/useVoiceAssistant.ts
bcherry: we probably need a full reference of special attributes but for now this is going into the avatar integration docs https://github.com/livekit/web/pull/1165/files#diff-42c10a5d6ec7f1e0d9ade709732ba1d4c3697b6d411523ecdc5a2ac75a9457efR46

lukasIO (approved):

PR #1136, opened 2025-05-12
Add examples for all useDataChannel usages to the first definition

bcherry:
~we [currently have two entries](https://github.com/livekit/livekit/issues/3617) in the table of contents that look identical. due to the design of our docs gen, there's no good solution without tradeoffs.~

~I think the most sensible thing to do is just document only one of them, with an example of usage without a topic. alternatively we could just get rid of the overrides and keep the one method that does both and has more complex type signatures and document that, but i think this approach here is best.~

Update: @lukasIO fixed the doc gen to just drop extra overloads so this PR is now just about adding all examples to the first one. https://github.com/livekit/components-js/pull/1138

### DIFF
--- a/packages/react/src/hooks/useDataChannel.ts
+++ b/packages/react/src/hooks/useDataChannel.ts
@@ -39,8 +39,6 @@ export function useDataChannel<T extends string>(
 ): UseDataChannelReturnType<T>;
 
 /**
- * Overload for `useDataChannel` without a topic. See {@link useDataChannel} for information and usage examples.
- *
  * @public
  */
 export function useDataChannel(
--- a/packages/react/src/hooks/useDataChannel.ts
+++ b/packages/react/src/hooks/useDataChannel.ts
@@ -13,7 +13,8 @@ type UseDataChannelReturnType<T extends string | undefined = undefined> = {
 
 /**
  * The `useDataChannel` hook returns the ability to send and receive messages.
- * By optionally passing a `topic`, you can narrow down which messages are returned in the messages array.
+ * Pass an optional `topic` to narrow down which messages are returned in the messages array.
+ *
  * @remarks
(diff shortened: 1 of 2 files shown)


packages/react/src/hooks/useDataChannel.ts
```diff
@@ -23,25 +24,24 @@ type UseDataChannelReturnType<T extends string | undefined = undefined> = {
...
- *
- * @example
- * ```tsx
- * // Send messages to all participants
- * const { message: latestMessage, send } = useDataChannel('chat', (msg) => console.log("message received", msg));
- * ```
- * @public
- */
+ * Overload for `useDataChannel` without a topic. See `useDataChannel<T>` for more usage examples.
+ * @internal
```
lukasIO, line 43: making it internal seems wrong though, that would suggest to not use this overload

bcherry: yeah i spent a while on this but can't find another way to mark it for doc exclusion, unfortunately

lukasIO: maybe @Ocupe has an idea wrt api extractor?

lukasIO: I think this should do it on the docs generation part: https://github.com/livekit/components-js/pull/1138

bcherry: perfect - once that's merged i'll adjust this so the one that gets documented will have the docstring describing both usages (unclear which one will be used from that PR, but it doesn't matter either way)

lukasIO: the one that's going to be used should always be the first one that's defined

bcherry: ok perfect - i just tested and it seems to work! updated

lukasIO (approved):

PR #1140, opened 2025-05-16
Bump livekit-client to 2.13.0

bcherry:
I want to ensure the data channel ordering change can get pulled through to react native so bumping this here so i can bump it there.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -7,8 +7,8 @@ settings:
 catalogs:
   default:
     livekit-client:
-      specifier: ^2.11.1
-      version: 2.11.1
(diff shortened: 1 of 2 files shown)


lukasIO:
similar situation for react native, it's declared as a peer dependency [here](https://github.com/livekit/client-sdk-react-native/blob/main/package.json#L87), so simply updating livekit-client independently should work for users, and if they're starting a new project it will automatically pull in the latest livekit-client version anyways.

I also opened https://github.com/livekit/components-js/pull/1143 which bumps livekit-client to support mediadevicekind in the mediadevicefailure callback.

bcherry:
sounds good, i'll consider all of these done then

PR #1177, opened 2025-07-02
Prevent crash when `settings` control is enabled in `ControlBar`

smorimoto:
The `ControlBar` component would crash when `settings: true` was set in the controls prop, despite being defined in the type definition. This occurred because the `SettingsMenuToggle` component was using `useLayoutContext()` which throws an error when no `LayoutContext` is provided.

Changed `useLayoutContext()` to `useMaybeLayoutContext()` in the `useSettingsToggle` hook to gracefully handle cases where `LayoutContext` is not available, preventing the crash and allowing the `settings` control to work properly.


lukasIO (approved):
thanks for the fix!

PR #1185, opened 2025-07-08
Fix `useParticipantTracks` not including local participant

mpnri:
Hi.
`useParticipantTracks` previously included local participant tracks, which make sense with the hook's name


mpnri:
Any chance for a review ?😅

mpnri:
@lukasIO 
image

PR #1188, opened 2025-07-16
fix(core): don't unregister stream handler on disconnect

lukasIO:
This ensures that a room instance can be reused and the transcription handler stays registered even after a disconnect.

### DIFF
--- a/packages/core/src/components/textStream.ts
+++ b/packages/core/src/components/textStream.ts
@@ -1,7 +1,7 @@
 import { RoomEvent, type Room } from 'livekit-client';
 import type { TextStreamInfo } from 'livekit-client/dist/src/room/types';
 import { from, scan, Subject, type Observable } from 'rxjs';
-import { share } from 'rxjs/operators';
+import { share, tap } from 'rxjs/operators';
 
 export interface TextStreamData {
   text: string;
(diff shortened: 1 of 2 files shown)


packages/core/src/components/textStream.ts
```diff
@@ -55,61 +55,71 @@ export function setupTextStream(room: Room, topic: string): Observable<TextStrea
...
-        // Handle case where stream ID wasn't found (new stream)
-        textStreams.push({
-          text: accumulatedText,
-          participantInfo,
-          streamInfo: reader.info,
+  // Create shared observable and store in cache
+  const sharedObservable = textStreamsSubject.pipe(
+    tap({
+      subscribe: () => {
+        room.registerTextStreamHandler(topic, async (reader, participantInfo) => {
```
lukasIO, line 66: registering logic is moved into `subscribe` logic of the textStreamsSubject, so we start listening only when we actually have some entity interested in it

packages/core/src/components/textStream.ts
```diff
@@ -55,61 +55,71 @@ export function setupTextStream(room: Room, topic: string): Observable<TextStrea
...
+
+              // Emit the updated array with the new stream
+              textStreamsSubject.next([...textStreams]);
+            }
+          });
         });
+      },
+      finalize: () => {
+        console.log('finalizing');
+        room.unregisterTextStreamHandler(topic);
```
lukasIO, line 109: removing the handler on finalize. This works because we also use `share` which does internal refcounting and finalize is only called once the refcount drops to 0

packages/core/src/components/textStream.ts
```diff
@@ -55,61 +55,70 @@ export function setupTextStream(room: Room, topic: string): Observable<TextStrea
...
   observableCache.set(cacheKey, sharedObservable);
 
   // Add cleanup when room is disconnected
-  room.once(RoomEvent.Disconnected, () => {
-    room.unregisterTextStreamHandler(topic);
-    textStreamsSubject.complete();
+  room.on(RoomEvent.Disconnected, () => {
     getObservableCache().delete(cacheKey);
+    textStreams = [];
+    textStreamsSubject.next([]);
```
lukasIO, line 121: these two lines simply reset the transcriptions to an empty array once the room disconnects. which is desired behaviour

PR #1190, opened 2025-07-17
fix(core): prevent audio output switching on safari based browsers

lukasIO:


packages/core/src/components/mediaDeviceSelect.ts
```diff
@@ -35,6 +36,11 @@ export function setupDeviceSelector(
         id === 'default' && localTrack.mediaStreamTrack.label.startsWith('Default') ? id : actualId,
       );
     } else if (room) {
+      const browser = getBrowser();
+      if (kind === 'audiooutput' && (browser?.name === 'Safari' || browser?.os === 'iOS')) {
+        log.warn(`Switching audio output device is not supported on Safari and iOS.`);
+        return;
+      }
```
1egoman, line 43: Is this the kind of thing it's worth throwing an error for because it should never happen?

lukasIO: Hm, if we were to throw an error here we would have to integrate some way to automatically detect feature support also in the MediaDeviceSelect component. 

I'm still not sure how to tackle this particular situation in the long term. We probably eventually want to support iOS, just that right now it's tricky to get the correct user-initiated permission for it to happen consistently. 

I'll leave this as is for now

PR #1197, opened 2025-08-04
Add new `useRoomConnection` hook for managing room connections

1egoman:
It seems initially like modeling a room connection with a `useEffect(() => { room.connect(); return () => room.disconnect(); }, [...])` would be natural and work well, however if dependencies in that `useEffect` change rapidly, then the room can thrash back and forth between connected and disconnected. Because `room.disconnect()` is async, this sometimes results in connects and disconnects happening at the same time, which can cause edge cases in the SDK, redulting in sometimes very weird errors due to state from past connection attempts being considered as current.

There's some thinking being done to address this at the client-sdk-js level but for now, introduce a hook folks can use to at least stem the bleeding in the short term a bit. Once the issue is fixed in `client-sdk-js` then this hook can be updated to just use a `useEffect` again.

Note that this hook inherently must make a tradeoff - the existing `room.connect()` / `room.disconnect()` logic only wraps part of each operation in a mutex, meaning that operations are supposed to be able to overlap to try to keep performance up at the expense of reliability. This hook takes the other side of that tradeoff, resulting in slower but much more predictable behavior if `connect()`s and `disconnect()`s occur rapidly.

Usage example: https://github.com/livekit-examples/agent-starter-react/pull/225

### DIFF
--- a/packages/react/src/hooks/useRoomConnection.ts
+++ b/packages/react/src/hooks/useRoomConnection.ts
@@ -2,18 +2,14 @@ import { Mutex, type TrackPublishOptions, Room } from 'livekit-client';
 import { useEffect, useMemo, useState } from 'react';
 import { useMaybeRoomContext } from '../context';
 
-/**
- * @public
- */
-type UseRoomConnectionDetails = {
+/** @public */
+export type UseRoomConnectionDetails = {
   serverUrl: string;
   participantToken: string;
 };
 
-/**
- * @public
- */
-type UseRoomConnectionOptions = {
+/** @public */
+export type UseRoomConnectionOptions = {
   getConnectionDetails: () => Promise<UseRoomConnectionDetails>;
   onConnectionError?: (err: Error) => void;
 
@@ -23,10 +19,8 @@ type UseRoomConnectionOptions = {
   trackPublishOptions?: TrackPublishOptions;
 };
 
-/**
- * @public
- */
-type UseRoomConnectionResult = {
+/** @public */
+export type UseRoomConnectionResult = {
   room: Room;
 
   /** What operation is the useRoomConnection hook currently in the midst of performing?  */
@@ -54,7 +48,7 @@ type UseRoomConnectionResult = {
 export function useRoomConnection(options: UseRoomConnectionOptions): UseRoomConnectionResult {
   const roomFromContext = useMaybeRoomContext();
   const room = useMemo(() => {
-    return roomFromContext ?? options.room ?? (new Room());
+    return roomFromContext ?? options.room ?? new Room();
   }, [options.room, roomFromContext]);
 
   const connected = options.connected ?? true;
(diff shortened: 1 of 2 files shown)


packages/react/src/hooks/useRoomConnection.ts
```diff
@@ -0,0 +1,97 @@
...
+  const connectDisconnectLock = useMemo(() => new Mutex(), []);
+  useEffect(() => {
+    connectDisconnectLock.lock().then(async (unlock) => {
+      if (connected) {
+        setStatus('connecting');
+        const connectionDetails = await options.getConnectionDetails();
+        try {
+          await Promise.all([
+            room.localParticipant.setMicrophoneEnabled(
+              true,
```
lukasIO, line 71: this limits the usage of this hook to scenarios where users connect to rooms with their mic (and only their mic) enabled 🤔

packages/react/src/hooks/useRoomConnection.ts
```diff
@@ -0,0 +1,97 @@
...
+  // NOTE: it would on the surface seem that managing a room's connection with a useEffect would be
+  // straightforward, but `room.disconnect()` is async and useEffect doesn't support async cleanup
+  // functions, which means `room.connect()` can run in the midst of `room.disconnect()`, causing
+  // race conditions.
+  const connectDisconnectLock = useMemo(() => new Mutex(), []);
+  useEffect(() => {
+    connectDisconnectLock.lock().then(async (unlock) => {
+      if (connected) {
+        setStatus('connecting');
+        const connectionDetails = await options.getConnectionDetails();
```
lukasIO, line 67: if we employ this pattern we have to think about what happens if retrieving connection details here fails. E.g. is a user expected to flip `connected` to false and back to true on an error? 

Also wondering if decoupling the two (connection and connectionDetails) makes more sense, as it can speed up the actual connection if connection Details retrieval happens in advance (e.g. like prefetch before you actually click)

packages/react/src/hooks/useRoomConnection.ts
```diff
@@ -0,0 +1,97 @@
...
+          options.onConnectionError?.(error as Error);
+        } finally {
+          setStatus('idle');
+        }
+      }
+      unlock();
+    });
+  }, [connected]);
+
+  return { room, status };
```
lukasIO, line 96: returning the room here might lead to patterns that we don't really want to encourage, i.e. trying to call methods on the room itself again.

1egoman: Ah, I had figured that was kinda the whole point, I'm not too well versed with this package yet. It sounds like the intention maybe is that the room would always be in the `RoomContext` and that is how other operations should be performed on it, including imperative updates?

packages/react/src/hooks/useRoomConnection.ts
1egoman: Yea, you are correct, but this doesn't seem bad to me / was intentional.

I suppose maybe it could be nice to be able to configure whether this `setMicrophoneEnabled` runs here or not, to enable / disable the audio buffering behavior before connection? But besides that, is there a scenario where a user would want to connect to a room with something else like video enabled before `room.connect(...)` runs?

PR #1198, opened 2025-08-05
Update `LiveKitRoom` connection logic

1egoman:
Initially, I made https://github.com/livekit/components-js/pull/1197, which aimed to create a react hook for managing a room connection. However, after some comments were left and I started digging in to the codebase a bit deeper, I noticed that there was already a mechanism in place to support this - `LiveKitRoom` / `useLiveKitRoom`. Embarrassingly, this is also called out right in the README... 🤦 

So, ditch the old attempt and instead update this pre-existing abstraction to use the connection behavior from the previous change. In addition, it looks like there isn't a means to support the microphone preconnect audio buffer logic with the existing implementation, so add a mechanism to do that.

I ported `agent-starter-react` over to use this mechanism here: https://github.com/livekit-examples/agent-starter-react/pull/225

### DIFF
--- a/packages/react/src/components/LiveKitRoom.tsx
+++ b/packages/react/src/components/LiveKitRoom.tsx
@@ -79,6 +79,12 @@ export interface LiveKitRoomProps extends Omit<React.HTMLAttributes<HTMLDivEleme
   simulateParticipants?: number | undefined;
 
   /**
+   * Optional side effect to run in parallel with the connection process.
+   * This can be used to enable the microphone preconnect buffer / etc
+   */
+  connectionSideEffect?: (room: Room) => Promise<void>;
+
+  /**
    * @internal
    */
   featureFlags?: FeatureFlags;
--- a/.changeset/pink-paths-pick.md
+++ b/.changeset/pink-paths-pick.md
@@ -1,0 +1,5 @@
+---
+'@livekit/components-react': patch
+---
+
+Update LiveKitRoom connection logic to be more robust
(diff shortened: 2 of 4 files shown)


packages/react/src/components/LiveKitRoom.tsx
```diff
@@ -78,6 +78,12 @@ export interface LiveKitRoomProps extends Omit<React.HTMLAttributes<HTMLDivEleme
 
   simulateParticipants?: number | undefined;
 
+  /**
+   * Optional side effect to run in parallel with the connection process.
+   * This can be used to enable the microphone preconnect buffer / etc
+   */
+  connectionSideEffect?: (room: Room) => Promise<void>;
+
```
1egoman, line 86: This new prop / parameter allows a user to inject in a side effect to run in parallel with the `room.connect(...)` call - an example of its use in `agent-starter-react`:
```tsx
// Within a react component:
const enableMicrophonePreConnectBuffer = useCallback(async (room: Room) => {
  room.localParticipant.setMicrophoneEnabled(true, undefined, {
    preConnectBuffer: appConfig.isPreConnectBufferEnabled,
  });
}, []);

// (later on in the component)
return (
  <LiveKitRoom
    token="..."
    serverUrl="..."
    connectionSideEffect={enableMicrophonePreConnectBuffer}
  >
    {/* rest of application here */}
  </LiveKitRoom>
);
```

Note that I could see `enableMicrophonePreConnectBuffer` being extracted into the eventual agents sdk that has been discussed.

PR #1199, opened 2025-08-11
Add room connect disconnect hack

1egoman:
Per @lukasIO's and my discussion, adds a hook that can be used to work around `useEffect`s not waiting for a `room.disconnect()` to finish in the cleanup function before running a `room.connect()`.

I tested it out in `agent-starter-embed` and it seemed to do what was expected.

### DIFF
--- a/packages/react/src/hooks/useSequentialRoomConnectDisconnect.ts
+++ b/packages/react/src/hooks/useSequentialRoomConnectDisconnect.ts
@@ -10,12 +10,8 @@ const ROOM_CHANGE_WARNING_THRESHOLD_MS = 1000;
 
 /** @public */
 export type UseSequentialRoomConnectDisconnectResults<R extends Room | undefined> = {
-  connect: R extends undefined
-    ? typeof Room.prototype.connect | null
-    : typeof Room.prototype.connect;
-  disconnect: R extends undefined
-    ? typeof Room.prototype.disconnect | null
-    : typeof Room.prototype.disconnect;
+  connect: typeof Room.prototype.connect & (R extends undefined ? null : unknown);
+  disconnect: typeof Room.prototype.disconnect & (R extends undefined ? null : unknown);
 };
 
 /**
--- a/packages/react/etc/components-react.api.md
+++ b/packages/react/etc/components-react.api.md
@@ -1083,8 +1083,8 @@ export function useSequentialRoomConnectDisconnect<R extends Room | undefined>(r
 
 // @public (undocumented)
 export type UseSequentialRoomConnectDisconnectResults<R extends Room | undefined> = {
-    connect: R extends undefined ? (typeof Room.prototype.connect) | null : (typeof Room.prototype.connect);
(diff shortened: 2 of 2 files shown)


packages/react/src/hooks/useSequentialRoomConnectDisconnect.ts
```diff
@@ -0,0 +1,126 @@
...
+          case 'disconnect':
+            await room
+              .disconnect(...message.args)
+              .then(message.resolve)
+              .catch(message.reject);
+            break;
+        }
+      }
+    });
+  }, []);
```
lukasIO, line 87: without the `room` as a dependency here it wouldn't be possible for users to create/change a room after the initial render. 
Maybe we can incorporate a similar pattern where frequent updates to the room object would also trigger a warning, but it's possible for users to change/update a room before connecting?

packages/react/src/hooks/useSequentialRoomConnectDisconnect.ts
```diff
@@ -0,0 +1,126 @@
...
+ * // Connecting to a room:
+ * useEffect(() => {
+ *   connect();
+ *   return () => disconnect();
+ * }, [connect, disconnect]);
+ * ```
+ *
+ * @public
+ */
+export function useSequentialRoomConnectDisconnect(
```
lukasIO: it's a very descriptive name. 
Definitely can see why you chose the name, but if you can think of something a bit more concise I wouldn't mind 😅

1egoman: It's a really hard one to name... that one was the best I could come up with. I'll keep pondering it but definitely open to suggestions!

packages/react/src/hooks/useSequentialRoomConnectDisconnect.ts
1egoman: Ah oops, I think I just forgot that `room` dependency! Plus I think there may need to be some extra logic to cut over other state cleanly when that changes.

Let me see what I can come up with on that second point!

packages/react/src/hooks/useSequentialRoomConnectDisconnect.ts
```diff
@@ -0,0 +1,126 @@
...
+  room: Room,
```
1egoman: It might be good to for room to have some sort of inactive value (ie, `null` / `undefined`) so if a user did create a room later they didn't have to play games to conditionally run this hook.

1egoman: Done

packages/react/src/hooks/useSequentialRoomConnectDisconnect.ts
1egoman: Both of those should be implemented! Though instead of an explicit exception to the check before connecting, I sized the check so one room change before a connect shouldn't be an issue.

PR #1200, opened 2025-08-13
chore(react): deprecate legacy useTrackTranscriptions

lukasIO:


packages/react/src/hooks/useTrackSyncTime.ts
```diff
@@ -4,6 +4,7 @@ import { useObservableState } from './internal';
 
 /**
  * @internal
+ * @deprecated Use useTrackSyncTime instead
  */
 export function useTrackSyncTime(ref: TrackReferenceOrPlaceholder | undefined) {
```
1egoman: Is this saying that a function is deprecated, but that a user should use the same deprecated function instead? Maybe there's a different `useTrackSyncTime` somewhere this is referring to, but as is it's kinda confusing.

Something that could maybe help that confusion - there's a `@link` tsdoc directive ([more info](https://tsdoc.org/pages/tags/link/)) that could maybe be used to link to the correct implementation. If this seems like a good plan, I can make a ticket to do a pass and apply this everywhere?

lukasIO: uff, thanks for catching, that doesn't make any sense. It was used internally only anyways, the deprecated marker serves primarily as a reminder for us to remove it entirely for the next iteration

lukasIO: yeah, using the link directive would be great!

MirShayan:
TypeScript error in useTrackSyncTime: type mismatch between trackSyncTimeObserver and useObservableState

## Bug Description
TypeScript compilation error in `useTrackSyncTime.ts` hook.

## Error Message
```
TS2345: Argument of type 'number' is not assignable to parameter of type 
'{ timestamp: number; rtpTimestamp?: number | undefined; }'.
```

## Location
`node_modules/@livekit/components-react/src/hooks/useTrackSyncTime.ts`, line with `useObservableState(observable, ...)`

## Environment
- @livekit/components-react: 2.9.16
- @livekit/components-core: 0.12.11
- livekit-client: 2.16.0
- TypeScript: 5.9.3
- Node: 20.16.0

## Expected Behavior
Code should compile without TypeScript errors.

## Actual Behavior
TypeScript reports type mismatch between `trackSyncTimeObserver` return type and `useObservableState` expected parameter type.

## Possible Fix
The observable returned by `trackSyncTimeObserver` appears to emit `number` instead of `{ timestamp: number; rtpTimestamp?: number }`. Either:
1. Update `trackSyncTimeObserver` to emit the correct object shape, or
2. Map the observable before passing to `useObservableState`

## Workaround
Using `skipLibCheck: true` in tsconfig.json, but this is not ideal.

PR #1208, opened 2025-09-15
feat(react): allow specifying multiple device kinds in MediaDeviceMenu props

activescott:
feat(react): allow specifying multiple device kinds in MediaDeviceMenu props

This is an update to the MediaDeviceMenu react component to allow specifying an array of device kinds instead of a single kind.

- The menu already rendered a multi-kind menu when no kind was specified, there was just no way to specify >1 kind to render when specifying a kind. This PR leverages that same rendering.
- This should be API-compatible. The one type change was that the existing `MediaDeviceMenuProps` interface which is now a type alias for the prior props interface (now renamed `MediaDeviceMenuPropsSingleKind`) and the new `MediaDeviceMenuPropsMultiKind` props interface.

We're already using this in our app, and thought we'd share with the community. If something needs adjusted for you to accept let me know!

### DIFF
--- a/packages/react/src/prefabs/MediaDeviceMenu.tsx
+++ b/packages/react/src/prefabs/MediaDeviceMenu.tsx
@@ -3,8 +3,13 @@ import * as React from 'react';
 import { MediaDeviceSelect } from '../components/controls/MediaDeviceSelect';
 import type { LocalAudioTrack, LocalVideoTrack } from 'livekit-client';
 
-/** @public */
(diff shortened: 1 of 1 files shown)


lukasIO:
thanks for the PR! 

This makes sense as an addition. The API check fails as the exposed type is being changed, but as you correctly stated, the new type is backwards compatible, so I don't see any issue in adding this in. 

You'll have to run `pnpm run api-extractor` to update the API docs and commit them as part of the PR in order for the API check to succeed

kapilthakare-cyberpunk:
[Commenting to indicate this PR is being closed due to cleanup process]

PR #1222, opened 2025-10-22
Allow localtrack passing on BarVisualizer

lukasIO:
BarVisualizer should also accept audio tracks directly

### DIFF
--- a/packages/react/src/components/participant/BarVisualizer.tsx
+++ b/packages/react/src/components/participant/BarVisualizer.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { useBarAnimator } from './animators/useBarAnimator';
 import { useMultibandTrackVolume, type AgentState } from '../../hooks';
-import { isTrackReference, type TrackReferenceOrPlaceholder } from '@livekit/components-core';
+import { type TrackReferenceOrPlaceholder } from '@livekit/components-core';
 import { useMaybeTrackRefContext } from '../../context';
 import { cloneSingleChild, mergeProps } from '../../utils';
 import { LocalAudioTrack, RemoteAudioTrack } from 'livekit-client';
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "format:write": "prettier --write \"**/src/**/*.{ts,tsx,md}\"",
     "gen:docs": "turbo run gen:docs",
     "gen:docs:watch": "nodemon --watch \"tooling/api-documenter/src/**/*\" --watch \"packages/react/src/**/*\" --ignore \"packages/react/src/assets/**/*\" -e js,jsx,ts,tsx -x \"pnpm gen:docs\"",
-    "lint": "turbo run lint",
+    "lint": "turbo run lint -- --quiet",
     "preinstall": "npx only-allow pnpm",
     "ci:publish": "turbo run build --filter=./packages/* && pnpm gen:docs && changeset publish",
     "start:next": "turbo run start --filter=@livekit/component-example-next...",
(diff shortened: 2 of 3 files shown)


packages/core/etc/components-core.api.md
```diff
@@ -410,26 +425,39 @@ export const PIN_DEFAULT_STATE: PinState;
...
+export type ReceivedAgentTranscriptionMessage = ReceivedMessageWithType<'agentTranscript', {
+    message: string;
+}>;
+
+// @public (undocumented)
+export type ReceivedChatMessage = ReceivedMessageWithType<'chatMessage', ChatMessage & {
     from?: Participant;
-}
+    attributes?: Record<string, string>;
+}>;
```
1egoman, line 440: thought: It looks like some changes from #1207 didn;t get included in this file - sorry about that, my bad 😞 

However, I am a bit suprised that the ci step didn't check this? Maybe this is something to make a ticket on to look into, since it seems like this file isn't being checked like the associated `components-react`one is.

packages/react/src/components/participant/BarVisualizer.tsx
```diff
@@ -23,7 +24,9 @@ export interface BarVisualizerProps extends React.HTMLProps<HTMLDivElement> {
   state?: AgentState;
   /** Number of bars that show up in the visualizer */
   barCount?: number;
+  /** @deprecated use `track` field instead */
   trackRef?: TrackReferenceOrPlaceholder;
+  track?: TrackReferenceOrPlaceholder | LocalAudioTrack | RemoteAudioTrack;
```
1egoman, line 29: suggestion: Other components currently seem to favor the `trackRef` prop name rather than `track`. I think I understand why you did it here (as a stepping stone to migrating away from `TrackReference`) but this means this component has two similar props which could be used at the same time, resulting in a less clear end interface.

What about instead, making `BarVisualizerProps` actually have two versions in a discriminated union, one that is the pre-existing version with `trackRef` (plus adding the deprecation), and the other getting rid of `trackRef` and adding `track` instead? The nice thing about doing it that way is the fields are mutually exclusive.

packages/core/etc/components-core.api.md
lukasIO: we don't check the API compat on CI for the core module as that's treated as an internal package. 
Don't have a strong opinion on whether or not to change that.

packages/react/src/components/participant/BarVisualizer.tsx
lukasIO: It's somewhat cleaner that way, true. But the downside of that is that the props wouldn't be as easy to inspect anymore, right? 

I thought deprecating one of them is pretty clear?

packages/core/etc/components-core.api.md
1egoman: Ah, I see what you are saying, I guess that makes a certain amount of sense. Thinking about this in another way then - is there a good reason to even have the api-extractor tool run on the `core` package and generate this artifact if it won't be consumed externally?

packages/react/src/components/participant/BarVisualizer.tsx
1egoman: To be clear I don't really have a strong opinion on this, it just struck me reading this code that it would be nice if they were mutually exclusive since they really should never both be set and then once you've migrated over to the new `track` prop, it is as if the old `trackRef` prop no longer exists .

By "props wouldn't be as easy to inspect anymore" if you mean that the component props wouldn't point to a singular interface on a "go to definition" type operation and instead would be an expression of some sort - yes, that is a good point. I guess there are pros and cons to both approaches, so do whatever you think makes sense.

packages/core/etc/components-core.api.md
lukasIO: yeah, good point, both options sound fine to me!

PR #1223, opened 2025-10-23
Ensure multibandvolume returns 0 if no track is passed

lukasIO:


packages/react/src/hooks/useTrackVolume.ts
```diff
@@ -119,6 +119,7 @@ export function useMultibandTrackVolume(
 
   React.useEffect(() => {
     if (!track || !track?.mediaStream) {
+      setFrequencyBands((val) => val.fill(0));
```
1egoman: suggestion: Testing it out in a js console,`Array.prototype.fill` looks to be a mutative operation, so I think while this might work now, it is has the potential to result in weirdness in the future since there's no copy being done here.

Maybe consider something like this?

```suggestion
      setFrequencyBands((val) => val.slice().fill(0));
```

1egoman (approved):

PR #1224, opened 2025-10-23
Change useAgent to return `undefined` for nullable TrackReference values

1egoman:
This is being done to make them fit a bit better into the existing `trackRef` prop on a few components, which right now accepts `TrackReference | undefined`.

I think given the history of how things evolved with `TrackReference`s it makes sense to do this, but I think it's a rather unfortunate change to make since `null` and `undefined` semantically tend to mean different things in javascript ([more info](https://kettanaito.com/blog/the-difference-between-null-and-undefined), [other interesting read](https://medium.com/@stephenthecurt/a-brief-history-of-null-and-undefined-in-javascript-c283caab662e)) and this case to me is semantically much more of a "null" (explicit absence of a value) than an "undefined" (value was never set).

Maybe it's something to litigate further as part of the next major version, however.

### DIFF
--- a/packages/react/etc/components-react.api.md
+++ b/packages/react/etc/components-react.api.md
@@ -75,8 +75,8 @@ import { WidgetState } from '@livekit/components-core';
 
 // @public (undocumented)
 export type AgentCallbacks = {
-    [AgentEvent.CameraChanged]: (newTrack: TrackReference | null) => void;
-    [AgentEvent.MicrophoneChanged]: (newTrack: TrackReference | null) => void;
+    [AgentEvent.CameraChanged]: (newTrack: TrackReference | undefined) => void;
+    [AgentEvent.MicrophoneChanged]: (newTrack: TrackReference | undefined) => void;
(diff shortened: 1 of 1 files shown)


packages/react/src/hooks/useAgent.ts
```diff
@@ -468,8 +466,8 @@ export function useAgent(session?: SessionStub): UseAgentReturn {
           failureReasons: null,
 
           // Clear inner values if no longer connected
-          cameraTrack: null,
-          microphoneTrack: null,
+          cameraTrack: undefined,
+          microphoneTrack: undefined,
         };
```
1egoman, line 471: Just wanted to call out that I kept these `key: undefined` type entries here so that destructuring would continue to work coming out of `useAgent` - ie, `const { cameraTrack, microphoneTrack } = useAgent(session)`. Otherwise this pattern would break down because `cameraTrack` + `microphoneTrack` wouldn't always be in the return value.

PR #1225, opened 2025-10-23
Remove waiting for agent to connect as part of session.start

1egoman:
If this behavior is still requested by a user, it can be replicated by calling `await agent.waitUntilAvailable();` after `await session.start()`.

Because the call was made to make `Session` no longer agent specific, I think this change makes sense, though it does to me seem unfortunate that a user would have to do both of those steps when it is expected to be a relatively common path.

Maybe there's a middle ground like checking if an agent was dispatched as part of the token payload and THEN run this same check, but even that has some amount of magic to it which may not always be what people want in every scenario, so idk.


lukasIO:
I agree that checking for dispatch might actually be the most expected behaviour. It doesn't really solve the case where users dispatch agents outside of the token scope, but expecting an agent to appear if it's part of the token scope seems reasonable

PR #1226, opened 2025-10-24
Make useSession wait for agent only if an agent was dispatched

1egoman:
As an alternative to https://github.com/livekit/components-js/pull/1225, instead of calling `agent.waitUntilAvailable(...)` on each `session.start()` call, instead only call it if the participant token dispatches an agent.

### DIFF
--- a/packages/react/src/hooks/useSession.ts
+++ b/packages/react/src/hooks/useSession.ts
@@ -10,6 +10,7 @@ import {
   TokenSourceFixed,
   TokenSourceFetchOptions,
   RoomConnectOptions,
+  decodeTokenPayload,
 } from 'livekit-client';
 import { EventEmitter } from 'events';
 
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,6 +9,9 @@ catalogs:
     '@livekit/protocol':
       specifier: ^1.42.0
       version: 1.42.0
+    livekit-client:
+      specifier: ^2.15.14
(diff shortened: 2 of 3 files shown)


packages/react/src/hooks/useSession.ts
```diff
@@ -469,7 +476,9 @@ export function useSession(
       ]);
 
       await waitUntilConnected(signal);
```
xianshijing-lk, line 478: [question](https://pullpo.io/cc?l=question) 

without knowing enough context on the code, I wonder if we shoul do both waitUntilConnected() and await agent.waitUntilAvailable(signal); in some cases ? 

or it should be something like 
if (tokenDispatchesAgent && agent) {
        await agent.waitUntilAvailable(signal);
      } else { 
    await waitUntilConnected(signal);
}

also, should you guard agent ?

1egoman: > or it should be something like

`waitUntilConnected` will block until a the underlying room is connected, and `agent.waitUntilAvailable` will block until the agent has connected. One is dependent on the other, and while I guess technically the agent could never connect until the room connects first, this should always happen in a sequential fashion so I'm not sure that it really matters that much. So I guess I don't have a strong preference given both should behave identically, but maybe lean slightly towards what I have now.

> also, should you guard agent ?

No, this isn't required - `agent` is always set, but its internal state changes as the agent goes through the connection lifecycle.

PR #1228, opened 2025-10-30
Add explicit failure when agent disconnects from the room

1egoman:
This should hopefully make the agents sdk more robust to cases like a local agent server being terminated, a hosted agent leaving a call due to it being connected longer than the drain threshold, etc.

It still is an open question how we can / should disambiguate between purposeful agent disconnects and outright crashes. It doesn't seem like today there is enough metadata to do this, so right now I am assuming all non user triggered disconnects to be of the "crash" variety.

I think if the metadata to disambiguate this properly becomes available, then it would be worth making this difference clear somehow - maybe purposeful disconnects go into `disconnected` (and maybe there's an event that is emitted that a user can tap into and reset their app's connection state since right now that would be managed out of band), and crashes retain this `failed` status with the reason.

### DIFF
--- a/.changeset/spicy-heads-smoke.md
+++ b/.changeset/spicy-heads-smoke.md
@@ -1,0 +1,5 @@
+---
+'@livekit/components-react': patch
+---
+
+Add explicit failure when agent disconnects from the room


packages/react/src/hooks/useAgent.ts
```diff
@@ -354,6 +355,37 @@ export function useAgent(session?: SessionStub): UseAgentReturn {
     };
   }, [room]);
 
+  // If the agent participant disconnects in the middle of a conversation unexpectedly, mark that as an explicit failure
```
pblazej, line 358: Trying to rethink the disambiguation here, IMO this is fine for single agent use case (no "handoff", (re)joining, etc.)

pblazej (approved):
Based on the discussion(s), let's treat disconnection as unconditional failure for now 👍

lukasIO (approved):
lgtm

PR #1229, opened 2025-11-04
Sort session messages by first received timestamp, not "timestamp" field

1egoman:
@davidliu discovered a case where on react native, agents messages don't always render in chronological order. The reason why seems to be because agent transcription timestamps are generated outside of the client, and chat message timestamps are client generated ([see here](https://github.com/livekit/components-js/blob/09b09eb069d4e663e0331b5b10f0988fd9b317e0/packages/core/src/components/chat.ts#L163)). Because of this, if the phone and server timestamps have drifted substantially, it could be possible for the timestamps to not be directly comparable.

He suggested to me that I adopt a strategy that was used in the legacy transcriptions implementation, where the first time a transcription was  received, a client generated timestamp was logged, and _this_ was used to sort them instead of the arbitrary timestamp associated with the transcription. This pull request contains my attempt at porting this logic to the `useSessionMessages` hook.

Taking a step back, I'm still kinda skeptical of this approach. IMO, it would be better if there could be one cononical source of timestamps on the server end, but unless the SFU were to somehow add metadata to a message I'm not sure this would be possible right now. Maybe others who are more deeply familiar with the system have more thoughts on how this could work!

### DIFF
--- a/packages/react/src/hooks/useSessionMessages.ts
+++ b/packages/react/src/hooks/useSessionMessages.ts
@@ -112,9 +112,7 @@ export function useSessionMessages(session?: UseSessionReturn): UseSessionMessag
     return merged;
   }, [transcriptionMessages, chat.chatMessages]);
 
-  const messageFirstReceivedTimeMapRef = React.useRef(
-    new Map<ReceivedMessage['id'], Date>(),
-  );
+  const messageFirstReceivedTimeMapRef = React.useRef(new Map<ReceivedMessage['id'], Date>());
   const sortedReceivedMessages = React.useMemo(() => {
     const now = new Date();
     for (const message of receivedMessages) {


pblazej:
As discussed, I don't think that's the case for Swift, as we carry the initial timestamp info (in a way described here), there's no explicit sorting (`OrderedDictionary`) https://github.com/livekit/client-sdk-swift/blob/b4febe223e86fc350cae985af52fed49caf04cc0/Sources/LiveKit/Agent/Chat/Receive/TranscriptionStreamReceiver.swift#L137

packages/react/src/hooks/useSessionMessages.ts
```diff
@@ -124,13 +144,25 @@ export function useSessionMessages(session?: UseSessionReturn): UseSessionMessag
     }
   }, [receivedMessages]);
```
lukasIO: issue (non-blocking): probably makes sense to use `sortedReceivedMessages` instead of `receivedMessages` ?

packages/react/src/hooks/useSessionMessages.ts
```diff
@@ -124,13 +144,25 @@ export function useSessionMessages(session?: UseSessionReturn): UseSessionMessag
...
+    return receivedMessages.sort((a, b) => {
+      const aFirstReceivedAt = messageFirstReceivedTimeMap.get(a.id);
+      const bFirstReceivedAt = messageFirstReceivedTimeMap.get(b.id);
+      if (typeof aFirstReceivedAt === 'undefined' || typeof bFirstReceivedAt === 'undefined') {
+        return 0;
+      }
+
+      return aFirstReceivedAt.getTime() - bFirstReceivedAt.getTime();
+    });
+  }, [receivedMessages, messageFirstReceivedTimeMap]);
```
lukasIO: issue: because `receivedMessages` is in a memo and `messageFirstReceivedTimeMap` is set in a useEffect that depends on `receivedMessages` every update will currently trigger sorting to run twice

1egoman: I'm not sure I understand? The `useEffect` should run every time that a new `receivedMessages` state value is computed by the `useMemo` but I don't see any sort of cyclical dependency that would cause it to run multiple times per update.

Just as a test, I added a `console.log('effect was run');` log to [here](https://github.com/livekit/components-js/pull/1229/files/4e4742ab41f12b86b1591fe452fc1cfc6a607c6b#diff-354cce4cb7578d0da2ae13f1402cbba861d40b9fdae5fefaf541d8533d1a273bR119) and recorded the below video - are you able to see the behavior you are pointing out in it? From what I can see I don't see two logs in quick succession like I'd expect if the `useEffect` was running multiple times for each update.

https://github.com/user-attachments/assets/2a93e447-1def-4079-a3e4-50a3dee25904

lukasIO: the issue is not in that effect, but rather in the memo computation of the sorted array, if you log in the memo [here](https://github.com/livekit/components-js/blob/4e4742ab41f12b86b1591fe452fc1cfc6a607c6b/packages/react/src/hooks/useSessionMessages.ts#L148), you should be able to see the issue

davidliu (approved):

packages/react/src/hooks/useSessionMessages.ts
1egoman: Ah I see, I misunderstood - that is a tricky one. I think I was able to fix this by combining together the `useMemo` and `useEffect` into one step.

PR #1230, opened 2025-11-10
Fix `useSession` return value stability

1egoman:
In a comment, @thomasyuill-livekit had mentioned that `useSession`'s return value seemed to not be stable ([link](https://github.com/livekit-examples/agent-starter-react/pull/298#discussion_r2504903389)). I looked into it and this was because the options being passed into `tokenSource.fetch(...)` for configurable token sources were not stable inherintly because of the api being expiosed (they were passed in on each render as a hook parameter) and that was cascading downwards and causing the reference instability.

So, to fix this, I've reintroduced the `areTokenSourceFetchOptionsEqual` function which can be run against each new set of options that are received every hook run and compute a stable value, which in my testing leads to much fewer downstream `session` reference changes.

### DIFF
--- a/packages/react/src/hooks/useSession.ts
+++ b/packages/react/src/hooks/useSession.ts
@@ -522,8 +522,6 @@ export function useSession(
 
       let tokenDispatchesAgent = false;
       await Promise.all([
-        // FIXME: swap the below line in once the new `livekit-client` changes are published
-        // room.connect(tokenSource, { tokenSourceOptions }),
         tokenSourceFetch().then(({ serverUrl, participantToken }) => {
           const participantTokenPayload = decodeTokenPayload(participantToken);
           const participantTokenAgentDispatchCount =
@@ -559,8 +557,6 @@ export function useSession(
 
   const prepareConnection = React.useCallback(async () => {
     const credentials = await tokenSourceFetch();
-    // FIXME: swap the below line in once the new `livekit-client` changes are published
-    // room.prepareConnection(tokenSource, { tokenSourceOptions }),
     await room.prepareConnection(credentials.serverUrl, credentials.participantToken);
   }, [tokenSourceFetch, room]);
   React.useEffect(


packages/react/src/hooks/useSession.ts
```diff
@@ -154,6 +154,88 @@ type UseSessionCommonOptions = {
...
+        break;
+      default:
+        // ref: https://stackoverflow.com/a/58009992
+        const exhaustiveCheckedKey: never = key;
+        throw new Error(`Options key ${exhaustiveCheckedKey} not being checked for equality!`);
+    }
+  }
+
+  return true;
+}
```
1egoman, line 189: chore: This has been copied from here: https://github.com/livekit/client-sdk-js/pull/1733. I did this so I could test the change but IMO this copied code shouldn't be merged.

So, before this should be merged, get the above linked pull request deployed into a new version on npm, then delete this code and update it to be an import.

packages/react/src/hooks/useSession.ts
```diff
@@ -154,6 +154,88 @@ type UseSessionCommonOptions = {
...
+    setMemoizedTokenFetchOptions(unstableRestOptions);
+  }, [isConfigurable, unstableRestOptions]);
+
+  const tokenSourceFetch = React.useCallback(async () => {
+    if (isConfigurable) {
+      if (!memoizedTokenFetchOptions) {
+        throw new Error(
+          `AgentSession - memoized token fetch options are not set, but the passed tokenSource was an instance of TokenSourceConfigurable. If you are seeing this please make a new GitHub issue!`,
+        );
+      }
```
1egoman: thought: There might be a better way to word this error message. This should be an invariant that I want to make clear should never happen in actual use but isn't easily representable in another way.

(clarifying a bit further - I don't _think_ this can happen in actual use, but I'm not 100% sure reasoning through it in my head. So if somebody can make this happen down the line I would be curious to get the reproduction instructions!)

packages/react/src/hooks/useSession.ts
```diff
@@ -154,6 +154,88 @@ type UseSessionCommonOptions = {
...
+
+/** Internal hook used by useSession to manage creating a function that properly invokes
+ * tokenSource.fetch(...) with any fetch options */
+function useSessionTokenSourceFetch(
+  tokenSource: TokenSourceConfigurable | TokenSourceFixed,
+  unstableRestOptions: Exclude<UseSessionConfigurableOptions, keyof UseSessionCommonOptions>,
+) {
+  const isConfigurable = tokenSource instanceof TokenSourceConfigurable;
+
+  const [memoizedTokenFetchOptions, setMemoizedTokenFetchOptions] =
```
lukasIO: **suggestion (if-minor):** there's a nice [util library ](https://github.com/kotarella1110/use-custom-compare) that would allow us to use your compare function directly to compare the deps.

Internally it uses a ref for comparison, which I personally feel would be slightly preferable to the useEffect pattern in case you don't want to use that library directly. Main reason is that the useEffect will always only run on the subsequent tick and makes the flow a bit harder to reason about with multiple dependencies

1egoman: Sounds good - migrated to a ref instead!

lukasIO: Ah, that was a misunderstanding, I meant that the ref usage would make things cleaner because it means we can get rid of the useEffect altogether. 

If we need to keep the useEffect, then your previous solution works just as well. 
FWIW, I think using [this approach](https://github.com/kotarella1110/use-custom-compare/blob/master/src/useCustomCompareMemoize.ts#L32-L48) would allow us to get rid of the useEffect (which was my primary motivation to suggest this change)

1egoman:
Discussed with lukas offline - the concern he was raising was mostly about `session.start` / `session.end` not being stable, which I think some work can be done to keep those stable so something like `useEffect(() => { session.start(); return () => session.end(); }, [session.start, session.end])` could work. I will do this in a follow up change.

packages/react/src/hooks/useSession.ts
1egoman: Oops, I forgot to do this before merging 🤦 . I will do this as part of the [follow up change](https://github.com/livekit/components-js/pull/1230#issuecomment-3520686122).

Issue #1227, opened 2025-10-27
Agent state management issues in `useAgent` hook


frankfarzan:
### Select which package(s) are affected

\@livekit/components-react

### Describe the bug

Hi @1egoman, I'm playing around with Agent SDK changes (from #1207) and observed some issues with state tracking/reporting in `useAgent` (Overall I like the changes though).

## Summary

The `useAgent` hook has two bugs that cause incorrect state reporting when
agents disconnect or are missing:

1. **Agent disconnections mid-conversation don't trigger `failed` state** -
   Agents that disconnect after successful connection don't set failure reasons
2. **`listening` state can occur with no agent participant** - State shows
   `'listening'` when `agent.internal.agentParticipant` is `null`

---

### Problem

When an agent successfully connects initially but then disconnects
mid-conversation, the state doesn't transition to `failed` and no failure
reasons are provided.

### Expected vs Actual

**Expected:**

- `agent.state` should be `'failed'`
- `agent.failureReasons` should be set

**Actual:**

- `agent.state` remains as `'listening'`
- No failure reasons are set

### Root Cause

It seems the failure detection in `useAgentTimeoutIdStore` (lines 192-204) only
works during initial connection timeout. Once the agent successfully connects,
the timeout completes and there's no mechanism to detect subsequent
disconnections.

**File**: `components-js/packages/react/src/hooks/useAgent.ts`

- Lines 192-204: Timeout logic
- Lines 436-446: Timeout lifecycle

### Reproduction

```typescript
// Agent process is started.
await session.start();
// Agent connects and works fine
// ...
// Kill the agent process
// Expected: state = 'failed', failureReasons = ['Agent disconnected']
// Actual: state = 'connecting' (or previous), failureReasons = null
```

---

### Problem

`agent.state` can be `'listening'` while `agent.internal.agentParticipant` is
`null`, indicating the agent is listening when no agent actually exists.

### Expected vs Actual

**Expected:**

- State reflects actual agent presence
- `'listening'` state should only occur when an agent participant exists

**Actual:**

```typescript
agent.state === "listening";
agent.internal.agentParticipant === null; // Inconsistent!
```

### Root Cause

It seems the state calculation (lines 394-425) sets state to `'listening'` based
solely on `localMicTrack`:

```typescript
if (localMicTrack) {
  state = "listening";
  bufferingSpeachLocally = true;
}
```

This happens when the local microphone is published (buffering audio) but no
agent has joined the room yet.

[...]

PR #1231, opened 2025-11-10
Add new explicit session pre connect buffering state

1egoman:
Per a github issue discussion (https://github.com/livekit/components-js/issues/1227), separates out the preconnect buffer into its own distinct state separate from `listening` called `pre-connect-buffering`.

In addition, get rid of the previous `isAvailable` boolean state and replace it with a few new states:
- `isConnected` - this is true if the agent is actually connected
- `canListen` - this is true if user interaction should be allowed, either due to the agent being connected or the preconnect audio buffer being active
- `isPending` - this is true if the client is connecting to the agent or the agent is currently initializing itself
- `isFinished` - this is true if the system is in a terminal state (`disconnected` or `failed`)

### DIFF
--- a/packages/react/src/hooks/useAgent.ts
+++ b/packages/react/src/hooks/useAgent.ts
@@ -92,7 +92,7 @@ type AgentStateAvailable = AgentStateCommon & {
   isFinished: false;
 
   /** Is the agent currently connecting or setting itself up? */
-  isLoading: false;
+  isPending: false;
 
   cameraTrack?: TrackReference;
   microphoneTrack?: TrackReference;
@@ -117,7 +117,7 @@ type AgentStatePreConnectBuffering = AgentStateCommon & {
   isFinished: false;
 
   /** Is the agent currently connecting or setting itself up? */
-  isLoading: false;
+  isPending: false;
 
   cameraTrack?: TrackReference;
   microphoneTrack?: TrackReference;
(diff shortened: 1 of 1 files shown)


packages/react/src/hooks/useSession.ts
```diff
@@ -477,7 +477,7 @@ export function useSession(
 
       await waitUntilConnected(signal);
       if (tokenDispatchesAgent) {
-        await agent.waitUntilAvailable(signal);
+        await agent.waitUntilConnected(signal);
       }
```
1egoman, line 470: thought: I'm a bit conflicted on the session having a `waitUntilConnected` method in addition to the agent also having a `waitUntilConnected` method and them meaning different things.

Originally I was thinking I'd call this `isAgentConnected` but `agent.isAgentConnected` is sort of repetitive so I dropped `Agent`. Going back to `isAvailable` or `isReady` could be other options but both seem sufficiently vague to not necesarily be clearly different from `canInteract`.

packages/react/src/hooks/useAgent.ts
```diff
@@ -69,30 +74,50 @@ type AgentStateCommon = {
...
-  /** Is the audio preconnect buffer currently active and recording because the agent hasn't
-   * connected yet? */
-  isBufferingSpeech: false;
+  /**
+   * Is the agent ready for user interaction?
+   *
+   * Note that this may not mean that the agent is actually connected - the audio pre-connect
+   * buffer could be active and recording user input before the agent actually connects.
+   * */
+  canInteract: true;
```
lukasIO: **issue:** as mentioned in the notion doc, I think both wording in the doc comment and the name here is misleading for pre-connect

1egoman: Fixed in 1dab40a

PR #1241, opened 2025-12-01
Remove requirement for ReceivedChatMessage to have an explicit type key

1egoman:
In past versions of the `ReceivedChatMessage` type, there wasn't a `type` key included as part of the defined type, as it was NOT part of a larger union of other message types.

Therefore, as the [agent sdk features](https://github.com/livekit/components-js/pull/1207) were being developed, this type key was added in a fully backwards compatible way (ie, `type?: 'chatMessage'`, so the absence of the `type` key in legacy code still works). Then a number of weeks later, some refactoring was done as part of this agents sdk change before the initial pull request got merged which added the `ReceivedMessageWithType` utility type. This had the unintentional side effect of accidentally making the `type` key required in `ReceivedChatMessage`, resulting in an accidental breaking change.

So, this pull request fixes this oversight and makes `type` actually an optional key for this `ReceivedChatMessage` case, undoing the accidental breaking api change.

### DIFF
--- a/packages/core/etc/components-core.api.md
+++ b/packages/core/etc/components-core.api.md
@@ -432,8 +432,6 @@ export type ReceivedAgentTranscriptionMessage = ReceivedMessageWithType<'agentTr
     message: string;
 }>;
 
-// Warning: (ae-forgotten-export) The symbol "ReceivedChatMessageWithRequiredType" needs to be exported by the entry point index.d.ts
-//
 // @public (undocumented)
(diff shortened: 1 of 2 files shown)


.changeset/honest-needles-throw.md
```diff
@@ -0,0 +1,5 @@
+---
+'@livekit/components-core': patch
```
1egoman, line 2: Technically this isn't a patch change, but given that it's trying to fix an accidental breaking change, I think patch probably makes sense. But I'm open to making it something different if that would be preferred.

lukasIO (approved):

PR #1242, opened 2025-12-03
Update examples with useSession/useAgent hooks

pblazej:
Also, a little change to `useSession` for the camera track...

### DIFF
--- a/examples/nextjs/pages/agent.tsx
+++ b/examples/nextjs/pages/agent.tsx
@@ -16,13 +16,20 @@ import { MediaDeviceFailure, TokenSource } from 'livekit-client';
 import styles from '../styles/VoiceAssistant.module.scss';
 import { generateRandomUserId } from '../lib/helper';
 
-function SimpleVoiceAssistant() {
+function SimpleAgent() {
   const agent = useAgent();
+
+  useEffect(() => {
+    if (agent.state === 'failed') {
+      alert(`Agent error: ${agent.failureReasons.join(', ')}`);
+    }
+  }, [agent.state, agent.failureReasons]);
+
   return (
     <BarVisualizer
       state={agent.state}
       barCount={7}
-      trackRef={agent.microphoneTrack}
+      track={agent.microphoneTrack}
       style={{ width: '75vw', height: '300px' }}
     />
   );
@@ -40,15 +47,15 @@ const AgentExample: NextPage = () => {
     [params],
   );
   const [userIdentity] = useState(() => params?.get('user') ?? generateRandomUserId());
-  const [shouldConnect, setShouldConnect] = useState(false);
 
   const session = useSession(tokenSource, {
     roomName,
     participantIdentity: userIdentity,
   });
 
+  const [started, setStarted] = useState(false);
   useEffect(() => {
-    if (shouldConnect) {
+    if (started) {
       session.start().catch((err) => {
         console.error('Failed to start session:', err);
       });
@@ -58,11 +65,11 @@ const AgentExample: NextPage = () => {
       });
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [shouldConnect, session.start, session.end]);
+  }, [started, session.start, session.end]);
 
   useEffect(() => {
     if (session.connectionState === 'disconnected') {
-      setShouldConnect(false);
+      setStarted(false);
     }
   }, [session.connectionState]);
 
@@ -79,10 +86,10 @@ const AgentExample: NextPage = () => {
       <SessionProvider session={session}>
         <div className={styles.room}>
           <div className={styles.inner}>
-            {shouldConnect ? (
-              <SimpleVoiceAssistant />
+            {started ? (
+              <SimpleAgent />
             ) : (
-              <button className="lk-button" onClick={() => setShouldConnect(true)}>
+              <button className="lk-button" onClick={() => setStarted(true)}>
                 Connect
               </button>
             )}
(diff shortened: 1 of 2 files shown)


examples/nextjs/pages/clubhouse.tsx
```diff
@@ -2,52 +2,87 @@
...
+        })
+        .catch((err) => {
+          console.error('Failed to start session:', err);
+        });
+    } else {
+      session.end().catch((err) => {
+        console.error('Failed to end session:', err);
+      });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
lukasIO, line 58: thought(non-blocking): 😢

examples/nextjs/pages/clubhouse.tsx
```diff
@@ -2,52 +2,87 @@
...
+  useEffect(() => {
+    const handleMediaDevicesError = (error: Error) => {
+      const failure = MediaDeviceFailure.getFailure(error);
+      console.error(failure);
+      alert(
+        'Error acquiring camera or microphone permissions. Please make sure you grant the necessary permissions in your browser and reload the tab',
+      );
+    };
+
+    session.internal.emitter.on(SessionEvent.MediaDevicesError, handleMediaDevicesError);
```
lukasIO: thought: @1egoman maybe we should think about a cleaner high level helper for this. 

using the internal emitters in the examples doesn't feel great.

examples/nextjs/pages/e2ee.tsx
```diff
@@ -34,22 +36,67 @@ const E2EEExample: NextPage = () => {
...
-  room.setE2EEEnabled(true);
+  const tokenSource = useMemo(() => {
+    return TokenSource.endpoint(process.env.NEXT_PUBLIC_LK_TOKEN_ENDPOINT!);
+  }, []);
+
+  const session = useSession(tokenSource, {
+    roomName,
+    participantIdentity: userIdentity,
+    participantName: userIdentity,
+    room,
```
lukasIO, line 49: question: @1egoman is there anything ensuring that session is still "in charge" of the room connection cycle when users use this pattern?

examples/nextjs/pages/voice-assistant.tsx
```diff
@@ -1,81 +1,104 @@
...
+      state={agent.state}
       barCount={7}
-      trackRef={audioTrack}
+      trackRef={agent.microphoneTrack}
       style={{ width: '75vw', height: '300px' }}
     />
   );
 }
 
 const VoiceAssistantExample: NextPage = () => {
```
lukasIO: nit: would be nice to rename this to agent example both in the method names and file

examples/nextjs/pages/clubhouse.tsx
1egoman: There was a better way to do this proposed, `useEvents`, though it didn't get much discussion because I figured it was fairly non controversial - https://github.com/livekit/components-js/blob/main/packages/react/src/hooks/useEvents.ts#L5.

So the thinking is the above could be something like the below instead:
```tsx
useEvents(session, SessionEvent.MediaDevicesError, handleMediaDevicesError);
```

examples/nextjs/pages/e2ee.tsx
1egoman: Not currently, no. It's an "escape hatch" you can use to provide a custom room and if the caller modifies the room's state extensively out of band it will definitely cause things to break. This definitely needs to be documented better than it currently is.

examples/nextjs/pages/clubhouse.tsx
1egoman: Also a `thought:`, but maybe it could make sense to destructure the session to get those `start` / `end` values out? I think you had said a few weeks back that does silence the warning?

examples/nextjs/pages/audio-only.tsx
```diff
@@ -1,32 +1,75 @@
...
+    () => (typeof window !== 'undefined' ? new URLSearchParams(location.search) : null),
+    [],
+  );
   const roomName = params?.get('room') ?? 'test-room';
-  const [userIdentity] = useState(params?.get('user') ?? generateRandomUserId());
+  const [userIdentity] = useState(() => params?.get('user') ?? generateRandomUserId());
+
+  const tokenSource = useMemo(() => {
+    return TokenSource.endpoint(process.env.NEXT_PUBLIC_LK_TOKEN_ENDPOINT!);
+  }, []);
```
1egoman: suggestion: I think if at all possible, the examples (where it can be done) should define the `TokenSource` outside the component as a constant rather than in the component in a memoized fashion.

Making sure that the `TokenSource` doesn't change reference proved to be a challenge when Thom was porting over agent-starter-react to use these new apis, and I'd like to try to push people away from injecting local component state into `TokenSource`s (configurable token sources should include all the metadata somebody would need in `TokenSource.custom` cases) unless you are doing something truly weird like agent-starter-react does to handle sandbox environments properly.

examples/nextjs/pages/customize.tsx
```diff
@@ -1,72 +1,107 @@
...
+  const tokenSource = useMemo(() => {
+    return TokenSource.endpoint(process.env.NEXT_PUBLIC_LK_TOKEN_ENDPOINT!);
+  }, []);
+
+  const session = useSession(tokenSource, {
+    roomName,
+    participantIdentity: userIdentity,
+    participantName: userIdentity,
+    room,
+  });
```
1egoman: suggestion: I think you can get rid of the `room` parameter here since it is just `new Room()`?

examples/nextjs/pages/clubhouse.tsx
1egoman: Migrated all instances of `session.internal` (which all happened to be even listener related) over to use this `useEvents` pattern.

PR #1246, opened 2025-12-08
Add participant identity, name, and metadata to root useAgent() return value

1egoman:
- @davidzhao had surfaced the need for agent participant `identity` for interfacing with with existing lower level apis (it looked like his use case was RPC calls)
- Given this I think it makes sense to expose the `agentParticipant`'s `identity` on the root `useAgent()` return value
- When I got into this a bit further, I realized `useParticipantInfo()` was returning name + metadata alongside identity, so it seems like all three are probably equally important (especially since all three are [exposed in `TokenSourceFetchOptions`](https://github.com/livekit/client-sdk-js/blob/fe2851fda48750e129373493cf86f37e0ab9582d/src/room/token-source/types.ts#L59-L61))

### DIFF
--- a/packages/react/src/hooks/useAgent.ts
+++ b/packages/react/src/hooks/useAgent.ts
@@ -79,7 +79,7 @@ type AgentStateAvailable = AgentStateCommon & {
   state: 'listening' | 'thinking' | 'speaking';
   failureReasons: null;
 
-  /** The client's assigned identity, coming from the JWT token. */
+  /** The agent's assigned identity, coming from the JWT token. */
   identity: Participant['identity'];
   name: Participant['name'];
   metadata: Participant['metadata'];


packages/react/src/hooks/useAgent.ts
```diff
@@ -77,6 +79,11 @@ type AgentStateAvailable = AgentStateCommon & {
   state: 'listening' | 'thinking' | 'speaking';
   failureReasons: null;
 
+  /** The client's assigned identity, coming from the JWT token. */
+  identity: Participant['identity'];
```
davidzhao, line 83: is it the agent's identity? or the end-user's identity? the comment seems to be indicating the latter. (i.e. `client`).

assuming it's the agent's, i'd recommend changing the comment to be clear.

davidzhao (approved):
lg, just would suggest updating the comment

packages/react/src/hooks/useAgent.ts
1egoman: It is the agent identity - good point, updated.

PR #1248, opened 2025-12-09
feat: LiveKit OSS Shadcn setup

thomasyuill-livekit:
### Summary

This PR introduces the foundation for a new `@livekit/shadcn` package and updates the Storybook development environment to support it.

#### New `@livekit/shadcn` Package
- Initial package setup with TypeScript configuration and build tooling
- Base UI components from shadcn/ui:
  - `Button`, `ButtonGroup`, `Toggle`
  - `Select`, `Separator`, `Tooltip`
  - `Alert`, `Sonner` (toast notifications)
- Base UI components from ai-sdk elements:
  - `Conversation`, `Message`, `Shimmer`

#### Storybook Updates
- Upgraded React to v19
- Added `next-themes` integration with a new `withNextThemes` decorator for theme switching support
- Added `AgentSessionProvider` decorator for agent-related stories
- Added Tailwind CSS support
- Updated story imports and configurations

#### Workspace
- Updated pnpm workspace catalog dependencies

### DIFF
--- a/packages/shadcn/package.json
+++ b/packages/shadcn/package.json
@@ -1,4 +1,5 @@
 {
+  "private": true,
   "name": "@agents-ui",
   "version": "1.0.0",
   "description": "",
@@ -11,8 +12,13 @@
   "keywords": [],
   "author": "",
   "license": "Apache-2.0",
-  "dependencies": {
+  "peerDependencies": {
     "@livekit/components-react": "workspace:*",
+    "livekit-client": "catalog:",
+    "react": "^19.0.0",
+    "react-dom": "^19.0.0"
+  },
+  "dependencies": {
     "@livekit/protocol": "catalog:",
     "@radix-ui/react-select": "^2.2.6",
     "@radix-ui/react-separator": "^1.1.8",
@@ -23,12 +29,9 @@
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
     "jose": "^6.0.12",
-    "livekit-client": "catalog:",
     "livekit-server-sdk": "^2.13.2",
     "lucide-react": "^0.555.0",
     "motion": "^12.16.0",
-    "react": "^19.0.0",
-    "react-dom": "^19.0.0",
     "sonner": "^2.0.7",
     "streamdown": "^1.6.9",
     "tailwind-merge": "^3.3.0",
@@ -41,7 +44,7 @@
     "@types/node": "^22.0.0",
     "@types/react": "^19",
     "@types/react-dom": "^19",
-    "eslint": "^8.56.0",
+    "eslint": "catalog:",
     "eslint-config-next": "15.5.2",
     "eslint-config-prettier": "^10.1.5",
     "eslint-plugin-import": "^2.31.0",
--- a/examples/nextjs/package.json
+++ b/examples/nextjs/package.json
@@ -23,7 +23,7 @@
     "@types/node": "^18.19.14",
     "@types/react": "^18.2.55",
     "@types/react-dom": "^18.2.18",
-    "eslint": "^8.56.0",
+    "eslint": "catalog:",
     "eslint-config-next": "^12.3.4",
     "source-map-loader": "^4.0.2",
     "typescript": "5.4.2"
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -49,7 +49,7 @@
     "@microsoft/api-extractor": "^7.36.0",
     "@size-limit/file": "^11.0.2",
     "@size-limit/webpack": "^11.0.2",
-    "eslint": "^8.56.0",
+    "eslint": "catalog:",
     "eslint-config-lk-custom": "workspace:*",
     "size-limit": "^11.0.2",
     "tsup": "^8.0.0",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -83,7 +83,7 @@
     "@types/react": "^18.0.25",
     "@types/react-dom": "^18.0.8",
     "@vitejs/plugin-react": "^4.3.2",
-    "eslint": "^8.56.0",
+    "eslint": "catalog:",
     "eslint-config-lk-custom": "workspace:*",
     "jsdom": "^26.0.0",
     "react": "^18.2.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,6 +9,9 @@ catalogs:
     '@livekit/protocol':
       specifier: ^1.43.4
       version: 1.43.4
+    eslint:
+      specifier: ^8.56.0
+      version: 8.57.1
     livekit-client:
       specifier: ^2.16.0
       version: 2.16.0
@@ -182,7 +185,7 @@ importers:
         specifier: ^18.2.18
         version: 18.3.5(@types/react@18.3.18)
       eslint:
-        specifier: ^8.56.0
+        specifier: 'catalog:'
         version: 8.57.1
       eslint-config-next:
         specifier: ^12.3.4
@@ -228,7 +231,7 @@ importers:
         specifier: ^11.0.2
         version: 11.1.6(esbuild@0.24.2)(size-limit@11.1.6)
       eslint:
-        specifier: ^8.56.0
+        specifier: 'catalog:'
         version: 8.57.1
       eslint-config-lk-custom:
         specifier: workspace:*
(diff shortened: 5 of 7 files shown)


packages/shadcn/components/ai-elements/message.tsx
```diff
@@ -0,0 +1,367 @@
...
+import { cn } from '@/lib/utils';
+
+export type MessageProps = HTMLAttributes<HTMLDivElement> & {
+  from: UIMessage['role'];
+};
+
+export const Message = ({ className, from, ...props }: MessageProps) => (
+  <div
+    className={cn(
+      'group flex w-full max-w-[80%] flex-col gap-2',
```
1egoman, line 20: question: Is there a storybook story for components in this file or ones in the other in the `ai-elements` directory (`conversation.tsx` or `shimmer.tsx`)? Or are those expected to be part of the next part of the wider split up change?

packages/shadcn/components/ai-elements/message.tsx
```diff
@@ -0,0 +1,367 @@
+'use client';
+
+import type { ComponentProps, HTMLAttributes, ReactElement } from 'react';
+import { createContext, memo, useContext, useEffect, useState } from 'react';
+import type { FileUIPart, UIMessage } from 'ai';
```
1egoman, line 5: suggestion: I am not exactly sure the full context behind why these types are being used from the `ai` package, but there's a message model that is being generated by the `useSessionMessages` hook (and also `useChat` before that) [which can be found here](https://github.com/livekit/components-js/blob/7790393773e71d640544f3794f5b72936e3aea42/packages/core/src/messages/types.ts#L47-L50) which I wanted to make you aware of if you didn't already know about it.

If the goal is to make something that works in a plug and play fashion with these new abstractions, it might be worth trying to integrate these models instead (though I don't think there's a 1:1 "role" type field available, and I don't think the "attachments" fields that exist are actively being used for anything).

packages/shadcn/components/ui/alert.tsx
```diff
@@ -0,0 +1,59 @@
+import * as React from 'react';
+import { type VariantProps, cva } from 'class-variance-authority';
+import { cn } from '@/lib/utils';
+
+const alertVariants = cva(
+  'relative w-full rounded-lg border px-4 py-3 text-sm grid has-[>svg]:grid-cols-[calc(var(--spacing)*4)_1fr] grid-cols-[0_1fr] has-[>svg]:gap-x-3 gap-y-0.5 items-start [&>svg]:size-4 [&>svg]:translate-y-0.5 [&>svg]:text-current',
+  {
+    variants: {
```
1egoman, line 8: question: How does the shadcn publishing process know to not include these components in the `ui/` directory? I'm assuming these are just copies of the pre-existing / stock ui components so that non atom level components can build on top of them, but maybe that is an incorrect assumption and these are expected to be published as well?

For context - I have only ever used shadcn as a user and never built / published my own component library, so it is very possible I'm just missing something obvious here.

packages/shadcn/package.json
```diff
@@ -0,0 +1,55 @@
...
+  "description": "",
+  "type": "module",
+  "main": "index.ts",
+  "scripts": {
+    "registry:build": "shadcn build --output ./dist",
+    "registry:serve": "python3 -m http.server 3210 -d dist"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
1egoman: question: Is this license correct? I don't think it matches with [the license at the root of the repository](https://github.com/livekit/components-js/blob/main/LICENSE).

packages/shadcn/package.json
```diff
@@ -0,0 +1,55 @@
...
+  "dependencies": {
+    "@livekit/components-react": "workspace:*",
```
1egoman, line 16: question: This is another question that might be just something I am missing, but how do you tell shadcn what version of `@livekit/components-react` that each component requires? Does it just assume the latest version?

1egoman (approved):
Generally makes sense to me!

packages/shadcn/components/ai-elements/message.tsx
thomasyuill-livekit: These are components installed from https://ai-sdk.dev/elements that are going to be dependencies of our `agents-ui` components. (eg. https://ai-sdk.dev/elements/components/message)

I do not plan to create stories for components that are dependencies, and not developed/maintained by us

1egoman: Ah, I didn't realize that! Ok, makes sense.

packages/shadcn/components/ai-elements/message.tsx
thomasyuill-livekit: I will be using the presentational `Message` component in our higher order `AgentChatTranscript` block component.

You can find an example of this in my exploration draft [PR](https://github.com/livekit-examples/agent-starter-react/pull/302), thought the block exposed by `agents-ui` is more generic.

thomasyuill-livekit: You can see this commit https://github.com/livekit/components-js/pull/1247/commits/13240e0d4f52cd694d1caf91e43a2a28813d029a for my future proposal

packages/shadcn/components/ui/alert.tsx
thomasyuill-livekit: the registry will be generated from the file in this commit
https://github.com/livekit/components-js/pull/1247/commits/00ea3add14649e29aee41ece46db5745627e084c

packages/shadcn/package.json
thomasyuill-livekit: good catch, I think this is a `pnpm init` default?

thomasyuill-livekit: or one of the init commands I used to start this

packages/shadcn/package.json
thomasyuill-livekit: great question

it would default to latest unless otherwise defined
https://ui.shadcn.com/docs/registry/registry-item-json#dependencies

packages/shadcn/components/ui/alert.tsx
thomasyuill-livekit: then we run `pnpm registry:build` and the output is put in `./dist`
we will then need to copy paste (or automate) the output into our hosting directory TBD

see https://ui.shadcn.com/docs/registry/getting-started#build-your-registry

tooling/docs-gen/package.json
```diff
@@ -25,7 +25,8 @@
     "eslint-config-lk-custom": "workspace:*"
   },
   "devDependencies": {
-    "@microsoft/api-extractor": "^7.35.2"
+    "@microsoft/api-extractor": "^7.35.2",
+    "eslint": "^8.56.0"
```
lukasIO, line 29: let's move this into the catalog as well as it seems to be set in all packages

packages/shadcn/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "@radix-ui/react-separator": "^1.1.8",
+    "@radix-ui/react-slot": "^1.2.3",
+    "@radix-ui/react-toggle": "^1.1.10",
+    "@radix-ui/react-tooltip": "^1.2.8",
+    "ai": "^5.0.105",
+    "class-variance-authority": "^0.7.1",
+    "clsx": "^2.1.1",
+    "jose": "^6.0.12",
+    "livekit-client": "catalog:",
+    "livekit-server-sdk": "^2.13.2",
```
lukasIO, line 32: the livekit dependencies should probably be peer dependencies?

packages/shadcn/components/ai-elements/conversation.tsx
lukasIO, line 1: is it correct to assume that we don't ever alter the components here in `/ai-elements` and `/ui` because we stick to the default implementations?

thomasyuill-livekit: correct, we wont alter them

but we will update them if/when shadcn publishes updates to these components

packages/shadcn/package.json
thomasyuill-livekit: Would this to make storybook more stable?

This package shouldn't ever be published or installed anywhere else

thomasyuill-livekit: I've just marked this package as "private" to indicate this

lukasIO: thanks for the callout, misinterpreted the dependencies

packages/shadcn/components/ai-elements/conversation.tsx
lukasIO: makes sense, I wonder if we should reflect that by not even checking them into git 🤔 
That would only be viable if there's a dev-ex friendly way to populate them on `pnpm i` though

thomasyuill-livekit: installing on post-install script would be a great idea
I'll make a note to do this in a future PR

1egoman: A post install build script sounds like a good idea. This was sort of what I was asking about [here](https://github.com/livekit/components-js/pull/1248?notification_referrer_id=NT_kwDOABoBLLMyMTA3NTU2ODc4OToxNzA0MjM2#discussion_r2604094573) but I think lukas phrased it better!

PR #1251, opened 2025-12-12
feat: LiveKit OSS Shadcn initial registry

thomasyuill-livekit:
## Summary

- Introduces a shadcn registry configuration (`registry.json`) for the `@agents-ui` package, enabling components to be installed via the shadcn CLI
- Restructures the agents-ui components to follow shadcn registry conventions by flattening nested component directories and consolidating hooks into a dedicated `hooks/agents-ui/` folder
- Consolidates `use-input-controls` and `use-publish-permissions` hooks into a single `use-agent-control-bar.ts` file
- Inlines `AgentChatInput` component into `agent-control-bar.tsx` to simplify the component structure

## Testing

- create a test next js app
- add components.json file (see below)
- run `pnpm registry:build && pnpm registry:serve`
- run `pnpm dlx shadcn@latest add @agents-ui/agent-track-toggle @agents-ui/agent-track-control @agents-ui/agent-disconnect-button @agents-ui/agent-control-bar @agents-ui/audio-visualizer-bar` in test app folder
- run `pnpm i && pnpm dev`
- try and use the `@agents-ui` components in your app

components.json
```
{
  "$schema": "https://ui.shadcn.com/schema.json",
  "style": "new-york",
  "rsc": true,
  "tsx": true,
  "tailwind": {
    "config": "",
    "css": "app/globals.css",
    "baseColor": "neutral",
    "cssVariables": true,
    "prefix": ""
  },
  "iconLibrary": "lucide",
  "aliases": {
    "components": "@/components",
    "utils": "@/lib/utils",
    "ui": "@/components/ui",
    "lib": "@/lib",
    "hooks": "@/hooks"
  },
  "registries": {
    "@agents-ui": "http://localhost:3210/{name}.json"
  }
}
```

### DIFF
--- a/packages/shadcn/registry.json
+++ b/packages/shadcn/registry.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://ui.shadcn.com/schema/registry.json",
   "name": "@agents-ui",
-  "homepage": "https://ui.livekit.io",
+  "homepage": "https://livekit.io/ui",
   "items": [
     {
       "name": "agent-disconnect-button",


packages/shadcn/components/agents-ui/agent-control-bar.tsx
```diff
@@ -1,21 +1,22 @@
-'use client';
```
lukasIO, line 1: in some places you're removing this directive in others you're adding it in this PR, can you share some context on when you want to use it and when not?

packages/shadcn/registry.json
```diff
@@ -0,0 +1,126 @@
...
+      "dependencies": [
+        "class-variance-authority",
+        "livekit-client",
+        "@livekit/components-react",
+        "lucide-react"
+      ],
+      "registryDependencies": ["@agents-ui/clone-single-child", "utils"]
+    },
+    {
+      "name": "clone-single-child",
```
lukasIO: is this something we want to have as part of the registry? 

To me it sounds like this would make more sense as an export from components-react

lukasIO:
@thomasyuill-livekit can you share some more info around the recommended flattening approach for hooks? 
The only official [example](https://github.com/shadcn-ui/registry-template/tree/main/registry/new-york/blocks/complex-component) I found nests the hook/lib/components folder in the component/block directory

packages/shadcn/registry.json
```diff
@@ -0,0 +1,126 @@
+{
+  "$schema": "https://ui.shadcn.com/schema/registry.json",
+  "name": "@agents-ui",
+  "homepage": "https://ui.livekit.io",
```
1egoman: question: This url doesn't seem to exist - what's going to be published to this url? The [shadcn docs](https://ui.shadcn.com/docs/registry/registry-json) seem to imply maybe a shadcn components specific registry landing page?

IMO, unless this is actively being worked on somewhere else, it would be worth replacing it with something else for now that is live (the docs site comes to mind as maybe a good interim option) and then swap this back in once this new url is live.

packages/shadcn/registry.json
```diff
@@ -0,0 +1,126 @@
...
+        {
+          "path": "components/agents-ui/agent-track-toggle.tsx",
+          "type": "registry:component"
+        }
+      ],
+      "dependencies": [
+        "class-variance-authority",
+        "livekit-client",
+        "@livekit/components-react",
+        "lucide-react"
```
1egoman, line 40: suggestion: Is there some way to specify at least a required major version for some of these dependencies, especially the livekit-controlled ones? Or even better, a semver range like "everything from the version where the relevant hooks / etc the given component relies on were introduced up until the next major version"?

If there were a new major version of a livekit controlled dependency, and shadcn always installs the `latest` of each one, it would be good for it to fail explicitly rather than implictly update to a new major version of the dependency and break all other older shadcn component versions / the app generally.

I bring it up because there have been discussions ongoing about a client sdk v3 initiative internally on the client team which would be a breaking change, so I don't think this necessarily is that farfetched of a scenario to happen in the next ~12 months.

packages/shadcn/registry.json
thomasyuill-livekit: Since we're just using this in the bar component I'm going to inline it.

Previously I was using it in the radial version too, but the DX was too clunky and I removed it

packages/shadcn/components/agents-ui/agent-control-bar.tsx
thomasyuill-livekit: While testing them in a sample app and only left them in for the ones that broke the app without them.

but I'm noticing that the ones I removed them in were wrapped in a client component that provides the required session object (through `SessionProvider`). So it's probably best I leave the `use client` in all of them.

I'll add them back in now

packages/shadcn/registry.json
thomasyuill-livekit: sure, can you suggest a sem version range for each as you seem more familiar with the requirements than I?

packages/shadcn/registry.json
thomasyuill-livekit: The plan is to get ui.livekit.io up ASAP (see [PR](https://github.com/livekit/web/pull/2604))
The first iteration will just serve the registry from this PR and we can iterate from there.
The soon we get ui.livekit.io up and running the soon we can register `agents-ui` on the open registry

I'm not sure this is of a big concern, elevenlabs registry still points to shadcn
https://ui.elevenlabs.io/r/registry.json

packages/shadcn/registry.json
1egoman: Copying from slack dm:
> I think for now, just pinning the major version of each would be good enough.
> Something more fancy could be done but I think that will give 90% of the benefit and should protect against new breaking livekit package changes

packages/shadcn/registry.json
thomasyuill-livekit: removed

packages/shadcn/components/agents-ui/agent-control-bar.tsx
thomasyuill-livekit: `use client` restored

packages/shadcn/registry.json
thomasyuill-livekit: We're going with https://livekit.io/ui for the homepage.
I've updated the registry to reflect this the web PR includes a placeholder landing page for now.

PR #1314, opened 2026-03-25
Add new useRpc hook prototype

1egoman:
### Overview 
A prototype of a new `useRpc` hook aiming to simplify both sending and responding to rpc messages. 

A snapshot build was published on npm, run the below to give it a shot locally:
```sh
$ npm install --save @livekit/components-react@rpc-prototype
```

### Example usage:

Here's the server end:
```tsx
// "Server" end of RPC assumes json encoded rpc requests / responses
useRpc(session, "getUserLocation", async (payload: { highAccuracy: boolean }, data) => {
  const position = await new Promise<GeolocationPosition>((resolve, reject) => {
    navigator.geolocation.getCurrentPosition(resolve, reject, {
      enableHighAccuracy: payload.highAccuracy,
      timeout: data.responseTimeout * 1000,
    });
  });
  return {
    latitude: position.coords.latitude,
    longitude: position.coords.longitude,
  };
});

// Alternative: custom request / response serialization:
useRpc(
  session,
  "getUserLocation",
  async (payload, data) => {
    return "raw text";
  },
  // To go back to raw serialization, pass serializers.raw() explicitly here:
  { serializer: serializers.raw() },
);

// Alternative: custom request / response serialization:
useRpc(
  session,
  "getUserLocation",
  async (payload /* <= decoded from base64 */, data) => {
    return "text to base64 encode";
  },
  {
    // Fully custom serializer example (base64):
    serializer: serializers.custom({
      parse: (raw) => window.atob(raw),
      serialize: (base64) => window.btoa(base64),
    })
  }
);
```

And the client end:

```tsx
const { performRpc } = useRpc();

// The "Client" end of RPC also defaults to json:
const result = await performRpc({
  destinationIdentity: participantIdentity,
  method: 'getUserLocation',
  payload: { highAccuracy: true },
});

// Example of using a serializer in the client end as well:
const result = await performRpc({
  destinationIdentity: participantIdentity,
  method: 'getUserLocation',
  payload: "raw text",
}, serializers.raw());
```

### Todo
- [x] Add missing changeset
- [x] I can't get the api-extractor check to pass in ci (it passed locally) and can't figure out why... Debug this.
- [x] Update `livekit-client` dependency to be a version which contains the changes in https://github.com/livekit/client-sdk-js/pull/1885 to get the ci build to pass

### DIFF
--- a/packages/react/src/hooks/useRpc.ts
+++ b/packages/react/src/hooks/useRpc.ts
@@ -21,13 +21,18 @@ export type RpcMethodDescriptor<Input = string, Output = string> = {
 
 /** @beta */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type RpcMethod<Input = any, Output = any> = RpcRawHandler | RpcMethodDescriptor<Input, Output>;
+export type RpcMethod<Input = any, Output = any> =
+  | RpcRawHandler
+  | RpcMethodDescriptor<Input, Output>;
 
 /** @beta */
-export type PerformRpcDescriptor<Input = string, Output = string> = Omit<PerformRpcParams, 'payload'> & {
+export type PerformRpcDescriptor<Input = string, Output = string> = Omit<
+  PerformRpcParams,
+  'payload'
+> & {
   parse?: (raw: string) => Output;
   serialize?: (val: Input) => string;
-  payload: Input,
+  payload: Input;
 };
 
 /** @beta */
@@ -39,7 +44,6 @@ export type UseRpcOptions = {
   /** Only accept RPCs from this participant. Others receive UNSUPPORTED_METHOD. */
   from?: string | Participant;
 };
-
 
 /**
  * Namespace for RPC helpers.
@@ -72,13 +76,15 @@ export const rpc = (() => {
   /* Overload: payload mode (for performRpc) */
   function json<Input, Output>(value: RpcJsonParams<Input>): PerformRpcDescriptor<Input, Output>;
   function json<Input, Output>(
-    handlerOrValue: RpcJsonParams<Input> | ((payload: Input, data: RpcInvocationData) => Promise<Output>)
+    handlerOrValue:
+      | RpcJsonParams<Input>
+      | ((payload: Input, data: RpcInvocationData) => Promise<Output>),
   ): RpcMethodDescriptor<Input, Output> | PerformRpcDescriptor<Input, Output> {
     if (typeof handlerOrValue === 'function') {
       return {
         parse: (raw: string) => JSON.parse(raw),
         serialize: (val: unknown) => JSON.stringify(val),
-        handler: handlerOrValue as RpcMethodDescriptor<Input, Output>["handler"],
+        handler: handlerOrValue as RpcMethodDescriptor<Input, Output>['handler'],
       };
     }
 
@@ -114,7 +120,10 @@ function isUseSessionReturn(value: unknown): value is UseSessionReturn {
   );
 }
 
-async function resolveHandler<Input, Output>(method: RpcMethod<Input, Output>, data: RpcInvocationData): Promise<string> {
+async function resolveHandler<Input, Output>(
+  method: RpcMethod<Input, Output>,
+  data: RpcInvocationData,
+): Promise<string> {
   if (typeof method === 'function') {
     return method(data);
   }
@@ -138,7 +147,7 @@ async function resolveHandler<Input, Output>(method: RpcMethod<Input, Output>, d
     } catch (e) {
       throw RpcError.builtIn('APPLICATION_ERROR', `Failed to serialize RPC response: ${e}`);
     }
-  } else if (typeof result !== "string") {
+  } else if (typeof result !== 'string') {
     throw RpcError.builtIn(
       'APPLICATION_ERROR',
       `Failed to serialize RPC response: return value from handler function not string. Did you mean to include a "serialize" RpcMethod key?`,
@@ -183,10 +192,7 @@ export function useRpc(
   methods?: Record<string, RpcMethod>,
   options?: UseRpcOptions,
 ): UseRpcReturn;
-export function useRpc(
-  methods?: Record<string, RpcMethod>,
-  options?: UseRpcOptions,
-): UseRpcReturn;
+export function useRpc(methods?: Record<string, RpcMethod>, options?: UseRpcOptions): UseRpcReturn;
 export function useRpc(
   methodsOrSession?: Record<string, RpcMethod> | UseSessionReturn,
   optionsOrMethods?: UseRpcOptions | Record<string, RpcMethod>,
@@ -235,7 +241,10 @@ export function useRpc(
         const from = optionsRef.current?.from;
         const fromIdentity = typeof from === 'string' ? from : from?.identity;
         if (fromIdentity && data.callerIdentity !== fromIdentity) {
-          throw RpcError.builtIn('UNSUPPORTED_METHOD', `Method not available for caller ${data.callerIdentity}`);
+          throw RpcError.builtIn(
+            'UNSUPPORTED_METHOD',
+            `Method not available for caller ${data.callerIdentity}`,
+          );
         }
 
         // Resolve the latest handler from the ref
--- a/packages/react/src/hooks/useRpc.ts
+++ b/packages/react/src/hooks/useRpc.ts
@@ -1,0 +1,278 @@
+import * as React from 'react';
+import {
+  type Participant,
+  RpcError,
+  type RpcInvocationData,
+  type PerformRpcParams,
+} from 'livekit-client';
+
+import { useEnsureSession } from '../context';
+import type { UseSessionReturn } from './useSession';
+
+/** @beta */
+export type RpcRawHandler = (data: RpcInvocationData) => Promise<string>;
+
+/** @beta */
+export type RpcMethodDescriptor<Input = string, Output = string> = {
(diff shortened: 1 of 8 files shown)


1egoman:
Ok, I think this is finally in a good state to get some review eyes on. Lukas and I have talked about it fairly extensively though so I'm hoping it's fairly straightforward 🤞

packages/react/src/hooks/useRpc.ts
```diff
@@ -0,0 +1,282 @@
...
+ *
+ * // Manual serializer instances
+ * const a = serializers.raw(); // Serializer<string, string>
+ * const b = serializer.json<{ foo: string }, { bar: string }>(); // Serializer<{ foo: string }, { bar: string }>
+ * ```
+ *
+ * @beta
+ */
+export const serializers = (() => {
+  /**
```
1egoman: A thought that came to mind - maybe for futureproofing it would be worth moving the `Serializer` type and associated presets like `serializers.json()` / `serializers.raw()` / etc into `livekit-client` now so that in the future when `Serializer` is used in other places (with tighter integration into `livekit-client` core features) the import location of this won't have to be deprecated here and moved.

packages/react/src/hooks/useRpc.ts
```diff
@@ -0,0 +1,282 @@
...
+ * useRpc(session, "myMethod", async (payload) => payload, { serializer: serializers.raw() });
+ *
+ * // Manual serializer instances
+ * const a = serializers.raw(); // Serializer<string, string>
+ * const b = serializer.json<{ foo: string }, { bar: string }>(); // Serializer<{ foo: string }, { bar: string }>
+ * ```
+ *
+ * @beta
+ */
+export const serializers = (() => {
```
lukasIO: question: why did you opt for an IIFE instead of a plain object notation here?

suggestion: it would feel cleaner to me if there was a `base` fn declared outside that would be used by all. Otherwise changing `custom` in the future would have a direct influence on raw and json

packages/react/src/hooks/useRpc.ts
lukasIO: that's a good point. Probably worth doing now

packages/react/src/hooks/useRpc.ts
```diff
@@ -0,0 +1,282 @@
...
+  handler: RpcHandler<SerializerInput<S>, SerializerOutput<S>>,
+  options?: UseRpcOptions<S>,
+): UseRpcReturn;
+/** @beta */
+export function useRpc<S extends Serializer<any, any>>(
+  methodName: string,
+  handler: RpcHandler<SerializerInput<S>, SerializerOutput<S>>,
+  options?: UseRpcOptions<S>,
+): UseRpcReturn;
+export function useRpc(
```
lukasIO, line 101: suggestion: there should probably be additional overloads that take 
- only a session
- nothing

for cases where you want to `useRpc` only for the performRpc return value

packages/react/src/hooks/useRpc.ts
```diff
@@ -0,0 +1,282 @@
...
+          throw RpcError.builtIn('APPLICATION_ERROR', `Failed to parse RPC response: ${e}`);
+        }
+      } else {
+        return room.localParticipant.performRpc(params as PerformRpcParams);
+      }
+    },
+    [room],
+  );
+
+  return React.useMemo(() => ({ performRpc }), [performRpc]);
```
lukasIO: I'm still going back and forth in my mind about returning this as an object. 
I personally slightly lean towards returning the function directly for simplicity. 

I know it will be less extensible, but given that the functionality is really well scoped around RPCs, I think it would be fine?
I'm sure you had a similar debate in your head when deciding to do that, what made you opt for the object?

packages/react/src/hooks/useRpc.ts
1egoman: Good idea - I can do that!

packages/react/src/hooks/useRpc.ts
1egoman: Ah, I missed those, but that was the intention. Just added in 5c0152c7!

packages/react/src/hooks/useRpc.ts
1egoman: Really the primary argument I think is extensibility - I don't think I have a better concrete argument than that. That being said, I think that in itself still is a pretty strong argument. The one thing I keep coming back to as an example of additional state this could maybe return alongside could be a global "is sending" type state similar to the way `useSessionMessages` works - I could see that being pretty useful if your application is sending lots of rpc messages to show some sort of global "rpc loading" type state and not have to manage it all fully yourself.

Here's some ideas a LLM had which vary wildly in quality:
```
⏺ Looking at the current shape — a single performRpc plus the side effect of registering a handler — here are plausible future additions:                  
                                                
  Mutation-style state for outbound calls (react-query flavor)                                                                                             
  - isPending / isLoading, error, data for the most recent performRpc invocation                                                                           
  - lastError specifically as an RpcError so users can pattern-match on code                                                                               
                                                                                                                                                           
  Handler-side introspection                                                                                                                               
  - isRegistered: boolean — reflects the useEffect lifecycle; useful when methodName is conditional
  - lastInvocation?: RpcInvocationData — the most recent inbound call's metadata (caller identity, payload, requestId)                                     
  - invocationCount — how many times this handler has fired (and/or error count)                                      
  - callers: Set<string> — identities that have invoked this method during the hook's lifetime                                                             
                                                                                                                                                           
  Shape-aware helpers (leveraging the serializer generics already threaded through)                                                                        
  - A call(destinationIdentity, payload) convenience bound to the same methodName/serializer the hook registered — so a component that both registers "foo"
   and calls "foo" on a peer doesn't restate the method/serializer                                                                                         
  - callAll(payload) / broadcast(payload) — fan-out to all remote participants, returning Promise<Record<identity, Result | RpcError>>                     
                                                                                                                                                           
  Control surface                                                                                                                                          
  - unregister() — imperative counterpart to the effect-driven registration
  - abort() / an AbortSignal threaded into performRpc so callers can cancel in-flight RPCs before responseTimeout                                          
                             
  Typed proxy                                                                                                                                              
  - A remote proxy object where remote[identity].invoke(payload) type-checks against the registered handler's signature — pairs naturally with the
  Serializer<Input, Output> type params already in UseRpcOptions<S>                                                                                        
                                                                   
  The mutation-style state and the isRegistered / lastInvocation introspection feel like the most natural next steps — they'd match how useSessionMessages 
  and similar session hooks tend to expose both an action and observable state.   
```

I would say I'm not 100% opposed to your suggestion but also there's enough potential possibilities here where I'm hesitant to commit to not having this extension point.

lukasIO: makes sense, maybe then it's more of a naming thing, how do you like 
```
const rpc = useRpc();
rpc.send(...)
// or 
rpc.call(...)
// or
rpc.perform(...)
```

lukasIO: personally a slight preference towards `.call` but that doesn't align perfectly with the existing naming scheme of performRpc

1egoman: I think out of those three I prefer the last one just because it's closest to the terminology currently in use in the js sdk. But I think generally any of those I would be ok with!

That being said I would be interested in hearing more info on your concerns with `performRpc` vs `perform` - is the problem that a common variable name could be `rpc` and `rpc.performRpc` repeats?

lukasIO: Yeah, in my head the `.performRpc` naming breaks down if we expose additional things like `isPending`. Calling it `isPendingRpc` seems pretty redundant.

1egoman: Ah, I see - think the only minor concern I have with `call` is it might be mistaken for [`Function.prototype.call`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call). Curious your take on that.

I'll make the update though and see what I think.

packages/react/src/hooks/useRpc.ts
1egoman: Done in 636e0bb1

packages/react/src/hooks/useRpc.ts
1egoman: Cool, done - associated client-sdk-js changes here: https://github.com/livekit/client-sdk-js/pull/1885

packages/react/src/hooks/useRpc.ts
1egoman: Updated here: d93c2be. I think my general take is that patterns using `rpc.call` read\ fairly well, but I don't necessarily like `const { call } = useRpc()` quite as much, since `call` is a fairly generic term / it isn't really clear it relates to rpc. I think for that reason I am sort of leaning towards `perform`, since `rpc.perform` reads fairly well and if somebody sees `perform` in isolation at least there's some shared terminology with the main `client-sdk-js` local participant method (plus zero chance of confusion with the function prototype method).

However I don't really have a strong opinion here, if you can give me some more pros for `call` I think I could be convinced.

lukasIO: perform is also fine for me!

1egoman: Cool, I think I'll go with that for now, changes here: b3548ab

1egoman:
(rebased on top of latest `main`)

1egoman:
@lukasIO I believe all the comments we had both left should be addressed - any concerns with merging this as is? If not, can you give me a ✅ ?

lukasIO (approved):

PR #1317, opened 2026-04-10
Add encryption to use session

1egoman:
Adds the ability to pass an `encryption` parameter to `useSession` which allows for an easier path for configuring e2ee on a room.

```tsx
const worker = new Worker(new URL('livekit-client/e2ee-worker', import.meta.url));

const session = useSession(tokenSource, {
  agentName: appConfig.agentName,
  encryption: {
    enabled: true,
    key: "...",
    worker,
  },
});
```

I have updated `examples/nextjs/pages/e2ee.tsx` to use this new `encryption` parameter.

Additionally, while fixing this, I discovered a small bug introduced as part of https://github.com/livekit/components-js/pull/1309 - `session.start` was no longer a stable reference, so unfortunately because of this most of the examples were stuck in an infinite connect / disconnect loop 😬 . I converted the associated state value into a ref (which looks to be safe in this case) and now makes `session.start`reference stable, making the examples all work again.

### Todo
- [x] Get the api extractor ci check to pass. It passes for me locally, but doesn't pass in ci... I don't get it.

### DIFF
--- a/.changeset/dry-nights-sing.md
+++ b/.changeset/dry-nights-sing.md
@@ -1,0 +1,5 @@
+---
+'@livekit/components-react': patch
+---
+
+Adds new "encryption" field to useSession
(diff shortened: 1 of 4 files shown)


1egoman:
@lukasIO Given your updates, would you say this is now in a mergable state?

lukasIO (approved):

PR #1318, opened 2026-04-11
fix(shadcn): api-token-route missing room_config

thomasyuill-livekit:
## Issue

Related to https://github.com/livekit-examples/agent-starter-react/pull/331

```
POST /api/token 500 in 341ms
Error: cannot decode message livekit.RoomConfiguration from JSON: undefined
    at POST (app/api/token/route.ts:41:42)
```

## Solution

This is happens as the result of an empty `room_config` when `AGENT_NAME` is blank


davidzhao (approved):
lgtm

PR #1319, opened 2026-04-13
Enable encryption by default if options are passed on session

lukasIO:


packages/react/src/hooks/useSession.ts
```diff
@@ -360,40 +361,56 @@ export function useSession(
...
+    if (typeof encryptionKey === 'string' || encryptionKey instanceof ArrayBuffer) {
+      if (externalKeyProviderRef.current === null) {
+        externalKeyProviderRef.current = new ExternalE2EEKeyProvider();
+      }
+      externalKeyProviderRef.current.setKey(encryptionKey).catch((e) => log.error(e));
+      return externalKeyProviderRef.current;
+    } else {
+      return encryptionKey;
+    }
+  }, [encryptionKey]);
```
1egoman, line 381: question: I'm not sure I know enough about how E2EE works internally to be able to know this, but what's the benefit to caching `ExternalE2EEKeyProvider` like this when the key changes? I would think it wouldn't matter since the room is being recreated anyway when this changes?

packages/react/src/hooks/useSession.ts
```diff
@@ -360,40 +361,56 @@ export function useSession(
...
-    return new Room(roomOptions);
-  }, [roomFromContext, optionsRoom, encryptionEnabled, encryptionKey, encryptionWorker]);
+    return room;
+  }, [roomFromContext, optionsRoom, keyProvider, encryptionWorker]);
+
+  React.useEffect(() => {
+    () => {
+      room.disconnect();
+    };
+  }, [room]);
```
1egoman, line 413: question: What's up with this - I think as is currently this is a no-op since the function inside isn't being run? I think I get the intention (close the old room when the reference changes) - maybe you forgot a `return` in front of the inner function?

packages/react/src/hooks/useSession.ts
```diff
@@ -611,7 +628,7 @@ export function useSession(
   );
 
   const setEncryptionEnabled = React.useCallback(
-    async (enabled: boolean) => room.setE2EEEnabled(enabled),
+    async (enabled: boolean) => room.setE2EEEnabled(enabled).catch((e) => log.error(e)),
```
1egoman: question: What's up with this - shouldn't the errors get passed through to the caller, not logged and disgarded?

packages/react/src/hooks/useSession.ts
lukasIO: uff, yeah forgot the return, thank you!

packages/react/src/hooks/useSession.ts
lukasIO: makes sense, wanted to prevent unhandled promise rejections, but this is directly exposed to users. Will revert it

packages/react/src/hooks/useSession.ts
lukasIO: the idea is that changing the key would _not_ recreate the room but simply set the new key on the key provider if it already exists

1egoman (approved):

PR #1321, opened 2026-04-21
fix(react): recreate Krisp processor per track

Topherhindman:
## Summary

Noticed with our homepage agent in www that `useKrispNoiseFilter` currently throws `InvalidAccessError: Failed to execute 'connect' on 'AudioNode': cannot connect to an AudioNode belonging to a different audio context.` on the second session any time a consumer calls `Room.disconnect()` followed by a new `session.start()` without a page reload.

## Root cause

The hook stored the `KrispNoiseFilterProcessor` in `React.useState` and only created it once (guarded by `if (!krispProcessor)`), reusing the same instance across mic republishes. A `KrispNoiseFilterProcessor`'s internal audio graph is bound to the `AudioContext` it was `init()`-ed against, and `Room.disconnect()` closes that context. When the mic republishes on a fresh `AudioContext`, attaching the cached processor to the new track calls `AudioNode.connect()` on nodes from the old (closed) context — hence the error.

Duplicating the processor into React state while `LocalAudioTrack` already owns it was the underlying design mistake; the state copy had no mechanism to track when the real processor had been destroyed out from under it.

## Fix

Three commits:

1. **`fix(react): recreate Krisp processor per track`** — remove the duplicated state and treat `LocalAudioTrack.getProcessor()` as the source of truth. A single effect handles both attach and enable/disable sync:
   - If the current track already has the Krisp processor attached, just sync `setEnabled(shouldEnable)`.
   - Otherwise, if the filter is wanted on, create a fresh `KrispNoiseFilterProcessor` bound to this track's `AudioContext` and attach it.

   Because nothing is cached across tracks, mic republishes can't produce a stale-processor-from-a-closed-context scenario.

2. **`fix(react): detach Krisp processor on cancel`** — handle the cancel-mid-attach race where `setProcessor` resolves after the effect was cancelled (e.g. unmount during attach). The IIFE checks `cancelled` after each `await` and undoes the attach via `track.stopProcessor()` so the track is left as we found it.

3. **`fix(react): guard processor cast and detach on cleanup`** — guard the returned `processor` value with a `name === 'livekit-noise-filter'` check before casting (prevents misrepresenting a non-Krisp `TrackProcessor` set by another module), and add a separate cleanup effect keyed on the track that detaches the processor on track change or unmount. Uses identity (`track.getProcessor() === processor`) so it only detaches the instance this hook owns. Closes the gap where a caller-owned `trackRef` outlives the hook's component or swaps to a different live track.

## Test plan

- [x] Reproduced on www with the homepage agent: `session.start()` → `session.end()` → `session.start()` errors on second start.
- [x] With the patched build installed, three back-to-back start/end cycles in the same tab succeed without the error and with the filter active.
- [x] Toggle reuse path still works: `setNoiseFilterEnabled` flipping false → true does not destroy and recreate the processor.
- [x] `pnpm --filter @livekit/components-react build` / `lint` / `test` all pass (no new lint warnings).

### DIFF
--- a/packages/react/src/hooks/cloud/krisp/useKrispNoiseFilter.ts
+++ b/packages/react/src/hooks/cloud/krisp/useKrispNoiseFilter.ts
@@ -100,9 +100,19 @@ export function useKrispNoiseFilter(options: useKrispNoiseFilterOptions = {}) {
       const processor = KrispNoiseFilter(options.filterOptions);
       try {
         await track.setProcessor(processor);
-        if (cancelled) return;
+        // If the effect was cancelled while setProcessor was in flight, the
+        // processor landed on a track the hook no longer manages (e.g. the
+        // component unmounted with the track still alive via a caller-owned
+        // trackRef). Undo the attach so the track is left as we found it.
+        if (cancelled) {
+          await track.stopProcessor();
+          return;
+        }
         await processor.setEnabled(true);
-        if (cancelled) return;
+        if (cancelled) {
+          await track.stopProcessor();
+          return;
+        }
         setIsNoiseFilterEnabled(true);
       } catch (e: any) {
         setIsNoiseFilterEnabled(false);
--- a/packages/react/src/hooks/cloud/krisp/useKrispNoiseFilter.ts
+++ b/packages/react/src/hooks/cloud/krisp/useKrispNoiseFilter.ts
@@ -44,24 +44,17 @@ export function useKrispNoiseFilter(options: useKrispNoiseFilterOptions = {}) {
   const [isNoiseFilterPending, setIsNoiseFilterPending] = React.useState(false);
   const [isNoiseFilterEnabled, setIsNoiseFilterEnabled] = React.useState(false);
   let micPublication = useLocalParticipant().microphoneTrack;
-  const [krispProcessor, setKrispProcessor] = React.useState<
-    KrispNoiseFilterProcessor | undefined
-  >();
   if (options.trackRef) {
     micPublication = options.trackRef.publication;
   }
 
   const setNoiseFilterEnabled = React.useCallback(async (enable: boolean) => {
     if (enable) {
-      const { KrispNoiseFilter, isKrispNoiseFilterSupported } =
-        await import('@livekit/krisp-noise-filter');
+      const { isKrispNoiseFilterSupported } = await import('@livekit/krisp-noise-filter');
 
       if (!isKrispNoiseFilterSupported()) {
         log.warn('LiveKit-Krisp noise filter is not supported in this browser');
         return;
-      }
-      if (!krispProcessor) {
-        setKrispProcessor(KrispNoiseFilter(options.filterOptions));
       }
     }
     setShouldEnable((prev) => {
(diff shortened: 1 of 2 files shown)


packages/react/src/hooks/cloud/krisp/useKrispNoiseFilter.ts
```diff
@@ -73,37 +66,65 @@ export function useKrispNoiseFilter(options: useKrispNoiseFilterOptions = {}) {
...
+        await import('@livekit/krisp-noise-filter');
+      if (cancelled) return;
+      if (!isKrispNoiseFilterSupported()) {
+        setIsNoiseFilterPending(false);
+        return;
+      }
+      const processor = KrispNoiseFilter(options.filterOptions);
+      try {
+        await track.setProcessor(processor);
+        if (cancelled) return;
```
1egoman: question: Is it possible that `track.setProcessor` could run here, the effect could rerun (`cancelled` could be set to true), the bailout would occur, and if the effect didn't run again (maybe due to a higher level component unmount / etc), the track could be left with the processor set?

Is it worth doing something in the effect cleanup function to teardown the track processor and leave the track in the state that it started in?

1egoman (commented):
This generally makes sense to me, but I would like @lukasIO to give the final approval given I am not that familiar with this code.

packages/react/src/hooks/cloud/krisp/useKrispNoiseFilter.ts
Topherhindman: Ahh yeah! Good catch. This was a real race. I went back and forth on where to put the detach.

The constraint I ran into: at the moment the cleanup function fires, it can't reliably know whether `track.setProcessor` has actually landed, because the await may still be outstanding. The cleanup is what causes the bailout, it doesn't happen after it. Cleanup also can't `await stopProcessor()`, so the best it could do is fire and forget. And because cleanup runs on any dep change (including `shouldEnable` toggling), detaching unconditionally there would destroy the processor on every toggle and re-pay Krisp's wasm + model init on re-enable.

The spot that can do exactly what you described is the IIFE itself, after the `await track.setProcessor(processor)` resolves, since it knows the attach landed and can
`await track.stopProcessor()` cleanly. That's what the `if (cancelled) { await track.stopProcessor(); return; }` branches handle do. Same contract as "teardown on cancel, leave the track as you found it," just placed where the async state is legible.

The one case that isn't covered is "effect succeeded fully, then the component unmounts with the track still alive." For the default micPublication path i dont believe that can happen, as `LocalAudioTrack.stop()` destroys the processor during the room's own teardown, so the track can't outlive us. It could happen if a caller points `trackRef` at a track they own and keep alive past this hook's lifetime though. I looked at covering that with either a fire and forget detach in the cleanup function or a two effect split (attach lifecycle + enable sync), but both introduced worse problems. The cleanup function version broke toggle reuse by destroying the processor on every shouldEnable flip, and the two effect split regressed `isNoiseFilterPending` handling, made the returned processor a stale state value instead of the live `track.getProcessor()`, and created a `processorRequested` latch that reattached on future track swaps even while the filter was off.

Happy to revisit the caller owned trackRef case if you think it's common enough to justify the complexity.  Maybe as a follow up? Otherwise I think I'd leave this PR scoped to the race you caught.

1egoman: I think what you've done makes sense to me - I also get the sense this isn't common.

One other possible idea that came to mind just now - you could put the ` await stopProcessor()` in an unrelated effect which would have the benefit of at least cleaning it up when the component the hook is in unmounts (or that secondary effect could be given a different set of dependencies, unlinked to the set passed into the effect that calls `await startProcessor()`).

packages/react/src/hooks/cloud/krisp/useKrispNoiseFilter.ts
```diff
@@ -73,37 +66,75 @@ export function useKrispNoiseFilter(options: useKrispNoiseFilterOptions = {}) {
...
+      }
+    })();
+    return () => {
+      cancelled = true;
+    };
+  }, [shouldEnable, micPublication?.track, options.filterOptions]);
+
+  const processor =
+    micPublication?.track instanceof LocalAudioTrack
+      ? (micPublication.track.getProcessor() as KrispNoiseFilterProcessor | undefined)
```
lukasIO: issue: this cast assumes that no other processor has been set on the track by some other module.

bcherry (approved):

PR #1324, opened 2026-04-27
Add support for passing in RN e2eeManager

davidliu:
RN uses a separate E2EEManager to handle the encryption which needs to be passed in. This PR adds a separate typing and path to allow for this (see [E2EEOptions](https://github.com/livekit/client-sdk-js/blob/be55ae20c3908538808ee8acb60659b70185e66a/src/e2ee/types.ts#L205)).

### DIFF
--- a/packages/react/src/hooks/useSession.ts
+++ b/packages/react/src/hooks/useSession.ts
@@ -195,8 +195,12 @@ type UseSessionEncryptionOptions =
       /** An instance of the E2EE webworker, which must be constructed using your js build tool's
        * webworker construction mechanism. */
       worker: Worker;
+
+      e2eeManager?: undefined;
     }
   | {
+      key?: undefined;
+      worker?: undefined;
       /**
        * For React Native usage: Pass the e2eeManager obtained from the `useRNE2EEManager()` hook
        * in the `\@livekit/react-native` package.
--- a/packages/react/src/hooks/useSession.ts
+++ b/packages/react/src/hooks/useSession.ts
@@ -14,6 +14,7 @@ import {
   BaseKeyProvider,
   RoomOptions,
   ExternalE2EEKeyProvider,
+  type BaseE2EEManager,
 } from 'livekit-client';
 import { EventEmitter } from 'events';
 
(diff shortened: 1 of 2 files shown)


packages/react/src/hooks/useSession.ts
```diff
@@ -195,9 +196,17 @@ type UseSessionEncryptionOptions =
        * webworker construction mechanism. */
       worker: Worker;
     }
```
1egoman, line 200: nitpick: I think you might want `e2eeManager?: undefined` here as well to ensure that they are fully mutually exclusive? It might be pedantic though because in practice I'm not sure if you'd ever switch between these two modes given it sounds like each only works on one platform.

packages/react/src/hooks/useSession.ts
```diff
@@ -195,9 +196,17 @@ type UseSessionEncryptionOptions =
...
+  | {
+      /**
+       * For React Native usage: Pass the e2eeManager obtained from the `useRNE2EEManager()` hook
+       * in the `\@livekit/react-native` package.
+       */
+      e2eeManager: BaseE2EEManager;
+    }
```
1egoman, line 209: thought: I wonder if it makes sense to attempt to call `useRNE2EEManager` from within `useSession` purely for react native contexts so the public api could be a little similar. How have these types of things historically been handled? I do see that [krisp is an optional peer dependency](https://github.com/livekit/components-js/blob/00dfba89eb81a1f1f19ea4bc7ac03f130d0d121a/packages/react/package.json#L68-L72) so it looks like there's precedent for doing it that way as well.

1egoman (approved):

packages/react/src/hooks/useSession.ts
davidliu: Makes sense! 👍

packages/react/src/hooks/useSession.ts
davidliu: Hmm, the react-native SDK has a dependency on the components-react package, I think that might cause a circular dependency?

1egoman: Hmm, yea, good point. I think given that, trying to do this is going to be more trouble than its worth. I'm good with what you have now!

PR #1330, opened 2026-05-20
fix(react): forward track to setupDeviceSelector in useMediaDeviceSelect

### LINKED ISSUE #1315: useMediaDeviceSelect doesn't pass track to setupDeviceSelector, making setActiveMediaDevice ineffective in preview mode
### Bug Report

`useMediaDeviceSelect` calls `setupDeviceSelector(kind, roomFallback)` without passing the `track` argument, even though `track` is accepted as a prop.

**File:** `packages/react/src/hooks/useMediaDeviceSelect.ts` line 63

```typescript
const { className, activeDeviceObservable, setActiveMediaDevice } = React.useMemo(
  () => setupDeviceSelector(kind, roomFallback),  // track is not passed
  [kind, roomFallback, track],
);
```

`setupDeviceSelector` in `@livekit/components-core` expects 3 arguments: `(kind, room, localTrack)`. When `localTrack` is provided, it correctly calls `localTrack.setDeviceId()`. Without it, it falls back to `room.switchActiveDevice()`, which does nothing in preview mode (pre-room-connect) since `room` is a dummy `new Room()` instance.

### Expected fix

```diff
- () => setupDeviceSelector(kind, roomFallback),
+ () => setupDeviceSelector(kind, roomFallback, track),
```

### Impact

In any pre-join / preview UI that uses `useMediaDeviceSelect` with a `track` from `usePreviewTracks` / `createLocalTracks`, calling `setActiveMediaDevice` silently does nothing — users cannot switch microphone or camera before joining a room.

### Versions

- `@livekit/components-react`: 2.9.20
- `@livekit/components-core`: 0.12.13
- `livekit-client`: 2.18.0

nikhilgupta58:
Closes #1315

## Root cause

`useMediaDeviceSelect` keeps `track` in its `useMemo` dependency array but never forwards it to `setupDeviceSelector`.

`packages/react/src/hooks/useMediaDeviceSelect.ts` (line 62-65):

```ts
const { className, activeDeviceObservable, setActiveMediaDevice } = React.useMemo(
  () => setupDeviceSelector(kind, roomFallback), // track is a dep but never passed
  [kind, roomFallback, track],
);
```

`packages/core/src/components/mediaDeviceSelect.ts` defines a 3-arg signature and branches on the track:

```ts
export function setupDeviceSelector(
  kind: MediaDeviceKind,
  room: Room,
  localTrack?: LocalAudioTrack | LocalVideoTrack,
) {
  ...
  const setActiveMediaDevice = async (id, options = {}) => {
    if (localTrack) {
      await localTrack.setDeviceId(...);   // preview / pre-join path
    } else if (room) {
      await room.switchActiveDevice(...);  // no-ops on the dummy `new Room()`
    }
  };
}
```

In pre-join / preview UIs there is no connected room, so `roomFallback` is a dummy `new Room()`. Because `localTrack` is `undefined`, `setActiveMediaDevice` takes the `room.switchActiveDevice` branch, which does nothing on that dummy room. The result is that `setActiveMediaDevice` silently fails to switch the mic or camera before joining.

## The fix

```diff
-    () => setupDeviceSelector(kind, roomFallback),
+    () => setupDeviceSelector(kind, roomFallback, track),
     [kind, roomFallback, track],
```

`track` is already in the `useMemo` deps, so no dependency-array change is needed. This is purely additive forwarding of an already-tracked value.

## Proof (verify-first)

I reproduced the bug with a deterministic test BEFORE writing the fix.

The test renders `useMediaDeviceSelect({ kind: 'audioinput', track })` in preview mode (no room, no RoomContext, so `roomFallback` is a fresh `new Room()`) with a spy `track.setDeviceId`, calls `setActiveMediaDevice('mic-2')`, and asserts the track was switched. It keeps the real `setupDeviceSelector` (only the device enumeration and logging are mocked so jsdom does not touch `navigator.mediaDevices`).

RED on current code (`packages/react`):

```
 × forwards the local track so setActiveMediaDevice switches the device
 FAIL src/hooks/useMediaDeviceSelect.test.ts > useMediaDeviceSelect in preview mode > forwards the local track so setActiveMediaDevice switches the device
 AssertionError: expected "vi.fn()" to be called with arguments: [ 'mic-2' ]
 Number of calls: 0
   45|     expect(setDeviceId).toHaveBeenCalledWith('mic-2');
 Test Files  1 failed (1)
      Tests  1 failed (1)
```

GREEN after the fix:

```
 ✓ src/hooks/useMediaDeviceSelect.test.ts (1 test)
 Test Files  1 passed (1)
      Tests  1 passed (1)
```

Full suites stay green: `packages/react` 8 passed, `packages/core` 98 passed. `pnpm build` (all packages), lint, and format check are clean.

## Risk

None expected. The change only forwards a value that was already a memo dependency, and the preview-track code path in `setupDeviceSelector` already existed and was simply unreachable from this hook.

## Out of scope

While testing I noticed a separate UI-feedback quirk: `setupDeviceSelector` updates an internal `activeDeviceSubject` but returns the room-event-based `activeDeviceObservable`, so the active-item highlight does not reflect a preview-mode switch even after this fix. That is a distinct issue from #1315 (which is about the switch itself not happening) and is not addressed here.

### DIFF
--- a/.changeset/forward-track-to-device-selector.md
+++ b/.changeset/forward-track-to-device-selector.md
@@ -1,0 +1,5 @@
+---
+'@livekit/components-react': patch
+---
+
+Forward the local `track` to `setupDeviceSelector` in `useMediaDeviceSelect` so `setActiveMediaDevice` switches the microphone or camera in pre-join / preview mode (no connected room).
--- a/packages/react/src/hooks/useMediaDeviceSelect.test.ts
+++ b/packages/react/src/hooks/useMediaDeviceSelect.test.ts
@@ -1,0 +1,47 @@
+import { renderHook, act } from '@testing-library/react';
+import type * as ComponentsCore from '@livekit/components-core';
+import { afterEach, describe, expect, test, vi } from 'vitest';
+import { useMediaDeviceSelect } from './useMediaDeviceSelect';
+
+// Minimal subscribable so we avoid touching navigator.mediaDevices in jsdom.
+const noopObserver = { subscribe: () => ({ unsubscribe: () => {} }) };
+
+// Mock only the device enumeration + logging so jsdom does not touch navigator.mediaDevices.
+// The real setupDeviceSelector is kept so we exercise the actual track-forwarding behaviour.
+vi.mock('@livekit/components-core', async (importOriginal) => {
+  const actual = await importOriginal<typeof ComponentsCore>();
+  return {
+    ...actual,
+    createMediaDeviceObserver: () => noopObserver,
+    log: { info: vi.fn(), debug: vi.fn(), warn: vi.fn(), error: vi.fn() },
+  };
+});
+
+afterEach(() => {
+  vi.clearAllMocks();
+});
+
+describe('useMediaDeviceSelect in preview mode', () => {
+  test('forwards the local track so setActiveMediaDevice switches the device', async () => {
+    // A preview track as returned by usePreviewTracks / createLocalTracks.
+    const setDeviceId = vi.fn().mockResolvedValue(undefined);
+    const getDeviceId = vi.fn().mockResolvedValue('mic-2');
+    const mockLocalTrack = {
+      setDeviceId,
+      getDeviceId,
+      mediaStreamTrack: { label: 'Fake Microphone 2' },
+    } as unknown as Parameters<typeof useMediaDeviceSelect>[0]['track'];
+
+    // No room and no RoomContext => preview mode (roomFallback becomes new Room()).
+    const { result } = renderHook(() =>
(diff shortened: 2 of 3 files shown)


1egoman (approved):
Makes sense to me, thanks for the contribution!

Doing a bit of digging through history it looks like this `track` param wasn't omitted for an intentional purpose, it used to be there but its removal [looks fairly accidental](https://github.com/nikhilgupta58/components-js/commit/31a2b04ec36e9a3524804a2dbb7c2b663ea18b83#diff-246daab5d71d5f9448a651e07a59775cbdccae8b7863e9f289fab063686772d1L60).

PR #1346, opened 2026-06-18
Add proof of concept of transcription ordering fix

1egoman:
Previously, transcriptions were ordered by the first actual transcription data stream packet arrival time.

What this meant was that in a situation like the below:
1. User transcription data stream opened
2. Agent transcription data stream opened
3. Agent transcription first data stream packet received
4. User transcription first data stream packet received

The transcriptions would be ordered in `Agent` -> `User` order, not `User` -> `Agent`.

To fix this, change transcription ordering so it is based off of when transcriptions data streams are initially opened, and push all this ordering logic further up into the `components-core` package so it is more easily testable (and then both `useTranscriptions` AND `useSessionMessages` get this same fix). Finally, add some tests to assert both this behavior and the related clock offset behavior which @davidliu reported and was fixed in https://github.com/livekit/components-js/pull/1229.

Fixes #1280 (I am pretty sure, but would like to get confirmation from the reporter first)

## Todo
- [ ] Exercise this locally outside of a test context in an example app
- [ ] Verify with reporter this actually fixes the problem

### DIFF
--- a/.changeset/smooth-cobras-check.md
+++ b/.changeset/smooth-cobras-check.md
@@ -1,0 +1,6 @@
+---
+'@livekit/components-react': patch
+'@livekit/components-core': patch
+---
+
+Order transcriptions by data stream initial open time, NOT by first packet arrival time
--- a/packages/core/src/components/textStream.test.ts
+++ b/packages/core/src/components/textStream.test.ts
@@ -91,7 +91,7 @@ describe('setupTextStream ordering', () => {
 
   it('orders by first receipt of the segment, not by when its text finishes streaming (issue #1280)', async () => {
     const { room, open } = createMockRoom();
-    const emissions: TextStreamData[][] = [];
+    const emissions: Array<Array<TextStreamData>> = [];
     const sub = setupTextStream(room, TOPIC).subscribe((streams) => emissions.push(streams));
 
     // The user speaks first, so their transcription stream is opened first...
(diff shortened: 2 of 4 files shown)


lukasIO (approved):
generally makes sense to me!

For agent use cases, I believe in theory there's still a timing issue wrt speech timing vs agent transcript timing. The sender timing should be consistent for these cases anyways, so we gain little from the `firstReceivedTime` timestamp unless the textstream were to open before actual data is being sent. I don't think that's the case though: https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/voice/room_io/_output.py#L454-L464
we only `stream_text` once the transcription is available

1egoman:
@lukasIO - for context, In the [original issue](https://github.com/livekit/components-js/issues/1280), it seemed like the reporter mentioned that empty transcription chunks can get sent before the actual "content containing" transcriptions get sent. 

Reading that agents snippet you pasted though, it makes me wonder if maybe this isn't really right, and a better direction here maybe could be to keep with the old strategy of ordering transcriptions based on the first received transcription chunk, but ONLY start aggregating chunks once the first non empty chunk is received (I'm assuming non empty means `.length > 0`? But not 100% sure, I wasn't able to reproduce this - it might be an edge case specific to certain LLMs or something like that).

So that would look something like:
1. User transcription data stream opened
2. Agent transcription data stream opened
3. Agent transcription data stream packet received (`""`)
4. Agent transcription data stream packet received (`""`)
5. User transcription data stream packet received (`""`)
6. Agent transcription data stream packet received (`"foo"`)
7. User transcription data stream packet received (`"bar"`)

And then the transcriptions would get ordered `Agent` -> `User` because `foo` came before `bar`.

Thoughts on that?

PR #1347, opened 2026-06-18
Fix intersection observer

### LINKED ISSUE #1325: manageSubscription fails with useIntersectionObserver and GridLayout

\@livekit/components-react

### What I'm expecting

When` manageSubscription` is enabled on VideoTrack, track subscription should follow tile visibility consistently:
1. **subscribe** when the tile becomes visible
2. **unsubscribe** when the tile is no longer visible (including GridLayout page changes where tiles are removed from the DOM)

### What happens instead

`manageSubscription` is unreliable in two cases:
- after the usehooks-ts v3 migration I guess, `useIntersectionObserver` was not wired correctly, so visibility updates (`intersectionEntry?.isIntersecting`) were not consistently reflected
- with GridLayout pagination, tiles are unmounted when leaving the page, and no intersection “exit” event is fired, so unsubscribe may not happen as expected

### Reproduction

1. Render participant videos with `manageSubscription={true}` on **VideoTrack**.
2. Use both layouts:
    - **GridLayout** with enough participants to enable pagination
    - **CarouselLayout** with scroll
3. Add logs for:
    - `intersectionEntry?.isIntersecting`,
    - calls to `publication.setSubscribed(...)`
4. In CarouselLayout, scroll tiles in/out of view
5. In GridLayout, switch pages so tiles are removed from DOM

### Observed

`intersectionEntry?.isIntersecting` is not reliable (always false), so visibility-based subscription logic does not behave as expected.
In GridLayout pagination, tiles can unmount without a usable visibility transition, so unsubscribe is missed if only driven by intersection updates.

### Expected

`isIntersecting` should correctly reflect tile visibility changes.
With `manageSubscription`, tracks should subscribe when visible and unsubscribe when hidden or unmounted (including grid page changes).

### Logs

```shell

```

### System Info

```shell
Package: @livekit/components-react
Version: 2.9.20
Browsers: Chrome, Firefox
OS: Windows 1
Node.js: 24
```

### Severity

annoyance

### Additional Information

_No response_

1egoman:
The `usehooks-ts` v2 -> v3 migration released in `@livekit/components-react@2.3.4` (from 7cd73d75, so this has been broken since 2024 😬) changed the `useIntersectionObserver` API. In v3 the hook returns `{ ref, isIntersecting, entry }`, and the caller must attach the returned ref callback to the element being observed.

`VideoTrack` was never updated to take this into account:
- It ignored the returned ref, so the observer was attached to nothing - isIntersecting was stuck at its initial false forever.
- It passed `mediaEl.current` as root, but root is the scroll-container/viewport to measure against (default `null`), not the target element.

As a result, manageSubscription was not seeing actual visibility changes: tracks didn't subscribe when tiles became visible.

Core fix authored by @M-Blobby - I've cherry picked it over from a fork. Fixes https://github.com/livekit/components-js/issues/1325

## Todo
- [x] I haven't tested this fix in an example app, so do that and make sure it's not obviously broken in some way

### DIFF
--- a/.changeset/fix-manage-subscription-intersection-observer.md
+++ b/.changeset/fix-manage-subscription-intersection-observer.md
@@ -1,0 +1,5 @@
+---
+'@livekit/components-react': patch
+---
+
+Fix `manageSubscription` on `VideoTrack` not following tile visibility
--- a/packages/react/src/components/participant/VideoTrack.tsx
+++ b/packages/react/src/components/participant/VideoTrack.tsx
@@ -41,12 +41,25 @@ export const VideoTrack: (
   ) {
     const trackReference = useEnsureTrackRef(trackRef);
 
-    const mediaEl = React.useRef<HTMLVideoElement>(null);
+    const mediaEl = React.useRef<HTMLVideoElement | null>(null);
     React.useImperativeHandle(ref, () => mediaEl.current as HTMLVideoElement);
 
-    const intersectionEntry = useHooks.useIntersectionObserver({ root: mediaEl.current });
+    const { ref: intersectionRef, entry: intersectionEntry } = useHooks.useIntersectionObserver({
+      root: null,
+    });
 
     const [debouncedIntersectionEntry] = useHooks.useDebounceValue(intersectionEntry, 3000);
+
+    const managedPublicationRef = React.useRef<RemoteTrackPublication | null>(null);
+    const manageSubscriptionRef = React.useRef(false);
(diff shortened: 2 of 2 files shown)


lukasIO (approved):

1egoman:
Made a little test app to verify this works:


https://github.com/user-attachments/assets/8c2c58b4-1b42-4c25-8850-2a6635f36a5d

PR #1361, opened 2026-07-08
fix(shadcn): re-fetch labeled devices when track already acquired

thomasyuill-livekit:
## Issue

[DGN-650](https://linear.app/livekit/issue/DGN-650/fix-trackselector-bug)

`TrackDeviceSelect` didn't render device options after accepting the `getUserMedia` prompt when the track was already acquired elsewhere (e.g. via `session.start()`)

Also discovered that buttons aren't disabled + `!pressed` if no tracks are available.

## Solution

- Refactored `AgentTrackControl` to handle track fetching so its aware if tracks are available/unavailable
- Disabled toggle if tracks are not available
- Sets `pressed=false` if tracks are not available
- Fetches labeled devices automatically if track was already acquired elsewhere (e.g. `session.start()`)

## Testing

Open [/?path=/story/agents-ui-agenttrackcontrol--default ](https://components-js-storybook-5kld-git-ty-new-branch.staging.livekit.io/?path=/story/agents-ui-agenttrackcontrol--default) in incognito mode (or reset user media permissions)

- [x] TrackSelect appears after user permission is accepted
- [x] `AgentTrackControl` is disabled and not pressed if user permission are denied

### DIFF
--- a/docs/storybook/stories/agents-ui/AgentTrackControl.stories.tsx
+++ b/docs/storybook/stories/agents-ui/AgentTrackControl.stories.tsx
@@ -47,7 +47,7 @@ export default {
 export const Default: StoryObj<AgentTrackControlProps> = {
   render: (args: AgentTrackControlProps) => {
     const [isCameraPressed, setIsCameraPressed] = React.useState(true);
-    const [isMicrophonePressed, setIsMicrophonePressed] = React.useState(false);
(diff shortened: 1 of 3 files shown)


thomasyuill-livekit:
Closing in favor of #1362 — renaming this PR's head branch to `ty/fix-agent-track-control-bug` caused GitHub to auto-close this PR (twice), so I opened a fresh one from the renamed branch instead. Same commits, same description.

PR #1398, opened 2026-07-23
feat(shadcn): add downstream registry publish scripts

thomasyuill-livekit:
## Issue

No Linear ticket linked - this work wasn't tracked against a ticket.

## Overview

This adds maintainer tooling for publishing `@agents-ui` registry updates into downstream repos and hardens the flow so generated PRs are based on the latest downstream `main`, formatted with repo-local dependencies installed, and reviewed before anything is pushed.

## Summary of changes

| Area | Change |
|---|---|
| Downstream publish scripts | Adds `publish-agent-starter-react.ts`, `publish-livekit-web.ts`, and `publish-downstream.ts` to clone downstream repos, apply the freshly built registry, and open PRs |
| Latest downstream base | Each downstream clone explicitly fetches the latest `origin/main` and creates its publish branch from that fetched ref |
| Formatting before PRs | Generated downstream changes run `pnpm format` before diff detection/commit/PR creation |
| Dependency install before format | Fresh downstream clones install dependencies before formatting so repo-local binaries like `oxfmt` are available |
| Confirmation gate | Before pushing/opening a PR, the script prints repo, branch, title, body, and diff stat, then requires confirmation |
| Publish robustness | Adds cleanup/interrupt handling and avoids committing/pushing when no downstream diff is produced |
| Script rename and entry points | Renames `registry:*` scripts to `shadcn:*`, adds `pnpm shadcn:publish:all`, and wires the Turbo task as uncached/interactive |
| Safer deploy copy | Replaces shell copy/clean commands in `update.ts` with `fs` APIs and refuses unsafe registry clean targets |
| Build hooks | Adds dependency-build hooks for `packages/shadcn` tests and Storybook dev |
## Testing

- `pnpm --filter @livekit/agents-ui format`
- `node --experimental-strip-types -e "import('./packages/shadcn/scripts/publish-downstream.ts').then(()=>console.log('ok'))"`
- `pnpm --filter @livekit/agents-ui test`
- `pnpm --filter @livekit/agents-ui exec tsc --noEmit --pretty false` no longer reports the new script `.ts` import errors; it still reports pre-existing registry component/template type errors.

I did not run the actual publish scripts because they can push branches and open PRs.


1egoman (approved):
High level this makes sense, but I wonder if it would be better for these types of scripts to be run in ci rather than manually? From my understanding of reading this, if that was done, then it looks like it would be a fairly changesets like process where a pull request would get created and somebody would approve it before the deployment would happen.

Maybe there's a reason why this has to happen manually though, idk.

thomasyuill-livekit:
> High level this makes sense, but I wonder if it would be better for these types of scripts to be run in ci rather than manually? From my understanding of reading this, if that was done, then it looks like it would be a fairly changesets like process where a pull request would get created and somebody would approve it before the deployment would happen.
> 
> Maybe there's a reason why this has to happen manually though, idk.

That's a great call-out and where we should move towards.

This was a first attempt to automate the process for myself, so I could test and iron out the kinks
I think once I feel confident that this works reliably we can move this into CI 

Still working on this tho.
I'll let you know when I get this in a stable/reliable.

PR #1415, opened 2026-08-04
feat(shadcn): add useExpression hook

theomonnom:


thomasyuill-livekit (changes requested):
This is great @theomonnom 

We should include a storybook example for testing and evaluate the behaviour with an agent that has expression enabled

And we should add tests for the hooks

I'm happy to contribute these to the PR if that's helpful

thomasyuill-livekit:
@theomonnom I saw a response in slack on this thread, but it's not appearing in linear 🤷 <br>copy/pasta'd here so I can respond.

> Thanks [**@Thom Foolery**](<https://live-kit.slack.com/team/U08NAHXBV3K>) — pushed both in c0fecc8e.
>
> **Storybook**​ (`AgentExpressionAura.stories.tsx`), three stories:
>
> •  `Default` — the `mood` override as a control<br>•  `Palette` — every mood side by side, for judging whether the warm/cool split actually reads as a set<br>•  `LiveAgent` — drives off `useAgentExpression()` against a real session and prints both the raw provider label and the matched mood, which is the one to point at an expressive agent
>
> **Tests**​ — 51 cases in `tests/use-agent-expression.test.ts`, covering `matchMood`, `parseExpressionLabel` and the hook.
>
> On evaluating against a live expressive agent: I ran the `expressive_agent` example (livekit/agents#6698) through each provider first, and the corpus in the test file is the labels it actually emitted, not invented ones. That surfaced the thing worth flagging:
>
> •  ​**fishaudio**​ sticks to its closed set<br>•  ​**cartesia**​ documents six emotions but emitted `sympathetic`, `frustrated`, `joking/comedic`, `elated`, `amazed`, `contemplative` — it drifts outside its own vocabulary<br>•  ​**inworld**​ is free-form English: `"soft, with genuine care"`, `"bright, upbeat energy"`, `"gently curious, welcoming"`<br>•  ​**xai**​ has no expression/emotion tag at all, so it never publishes `lk.expression`
>
> So the label can't be switched on directly. The hook keyword-matches into a normalized mood, always exposes the raw `label`, takes `extraKeywords` to extend without forking the table, and falls back to `calm` on a miss. The corpus is a sample and the real variety is larger, so the matcher is deliberately best-effort rather than exhaustive.
>
> Very happy to have you contribute on top — particularly on the color palette, which is the piece I'd least trust myself on.
>
> One open question you may have a view on: there's guidance that it's risky to cement patterns into Agents UI visualizers before testing across models, and that agent-starter-react should get expressive support first. The hook and color presets are on that extraction list; a new visualizer isn't. I'm happy to downscope this PR to just the hook + presets and move the visualizer into the starter template if you'd prefer that ordering.

theomonnom:
@thomasyuill-livekit yeah somehow Claude code sent a comment so I removed it lol 

Otherwise I updated the code and added the storybook

thomasyuill-livekit:
lol

docs/storybook/stories/agents-ui/AgentExpression.stories.tsx
```diff
@@ -0,0 +1,198 @@
+import * as React from 'react';
+import { StoryObj } from '@storybook/react-vite';
+import { AgentSessionProvider } from '../../.storybook/lk-decorators/AgentSessionProvider';
+import {
+  AgentAudioVisualizerAura,
+  AgentAudioVisualizerBar,
+  AgentAudioVisualizerGrid,
+  AgentAudioVisualizerRadial,
```
thomasyuill-livekit: we forgot `AgentAudioVisualizerWave`

docs/storybook/stories/agents-ui/AgentExpression.stories.tsx
```diff
@@ -0,0 +1,198 @@
...
+};
+
+/**
+ * The end-to-end path against a real session: free-form provider label in, mood and color out.
+ *
+ * Leave the `mood` control empty to read the live agent. Setting it overrides the session, which
+ * is handy for checking a specific color without having to talk the agent into that mood.
+ */
+export const LiveAgent: StoryObj<UseExpressionOptions> = {
+  parameters: { layout: 'centered' },
```
thomasyuill-livekit:
```suggestion
  decorators: [LiveAgentSessionProvider],
  parameters: { layout: 'centered' },
```

You need a LiveAgentSessionProvider to hook this up to a real agent session, but we also need to configure the env vars to use a project/agent that has the expression API enabled

packages/shadcn/hooks/agents-ui/use-expression.ts
```diff
@@ -0,0 +1,578 @@
...
+    fallback,
+    session,
+  } = options;
+
+  const { messages } = useSessionMessages(session);
+
+  const { mood: liveMood, label } = useMemo(() => {
+    let found: string | null = null;
+    // last one wins; user turns never carry an expression
+    for (const message of messages) {
```
thomasyuill-livekit: feels like we could optimize this by keeping a reference to the last found message index with an expression, and then starting from that index in subsequent searches

thomasyuill-livekit (commented):
I really think the `useExpression` should focus on one task, 
extracting the normalized mood enum value and raw label

mapping it to a color, blending it with another color, etc are all implementation details of the identified mood. We can extract these into utility functions or hooks, or demonstrate in `agenter-starter-react`. Though they are useful for the purposed of storybook, I think they complicate and bloat the API

theomonnom:
@thomasyuill-livekit updated

packages/shadcn/hooks/agents-ui/use-expression.ts
```diff
@@ -0,0 +1,614 @@
...
+      } else if (label && segment.participantInfo.identity === speaker) {
+        // only the expressing participant's own later turns age the mood
+        turnsSince++;
+      }
+    }
+
+    const current = ttlTurns > 0 && turnsSince >= ttlTurns ? null : label;
+    const mood = current ? matchMood(current, { extraKeywords, fallback }) : null;
+
+    return { mood, label: current, color: moodColor(mood) };
```
thomasyuill-livekit: why are we still returning a color?
state and presentation should be separate concerns?

this hook's only job should be returning the current state of the agent's mood

thomasyuill-livekit (changes requested):
I believe we should add the `useExpression` hook to JS SDK and not Shadcn so it can be used by any React project.

our Shadcn Agents-UI library should be primarily concerned with presentation
since all the visualizers all ready support a color prop, I don't believe that Agents UI needs to be extended to support expressions.

We should not be including opionated "glue" code into our presentational components because that will make them brittle and less extensible. We should instead provide examples of how to glue these pieces together in our docs and starter templates

packages/react/src/hooks/useExpression.ts
```diff
@@ -0,0 +1,155 @@
...
+ * The expression published on a transcript segment, or null when it carries none.
+ */
+export function parseExpression(segment: TextStreamData): ExpressionPayload | null {
+  const raw = segment.streamInfo.attributes?.[EXPRESSION_ATTRIBUTE];
+  if (!raw) {
+    return null;
+  }
+
+  try {
+    const { value, mood } = JSON.parse(raw) as { value?: string; mood?: AgentMood };
```
thomasyuill-livekit: are we now normalizing the expression to `mood` in the Agents SDK?

packages/react/src/hooks/useExpression.ts
```diff
@@ -0,0 +1,155 @@
...
+    let ticks = 0;
+    const id = setInterval(() => {
+      setSettled((previous) => {
+        let latest: string | null = null;
+        for (const segment of segments) {
+          const parsed = parseExpression(segment);
+          if (parsed) {
+            latest = parsed.label;
+          }
+        }
```
thomasyuill-livekit: why wouldn't we go backwards through the array? 

exit as soon as we find the first parsed expression and compare it with the previous

```suggestion
        for (const segment of segments.reverse()) {
          const parsed = parseExpression(segment);
          if (parsed) {
            latest = parsed.label;
            return latest === previous ? previous : latest;
          }
        }
```

packages/react/src/hooks/useExpression.ts
```diff
@@ -0,0 +1,155 @@
...
+      const parsed = parseExpression(segment);
+      if (parsed) {
+        current = parsed;
+        speaker = segment.participantInfo.identity;
+        turnsSince = 0;
+      } else if (current && segment.participantInfo.identity === speaker) {
+        // only the expressing participant's own later turns age the mood
+        turnsSince++;
+      }
+    }
```
thomasyuill-livekit: again couldn't we do this in reverse order?

```suggestion
    turnsSince = 0;
    for (const segment of segments.reverse()) {
      turnsSince++;
      const parsed = parseExpression(segment);
      if (parsed) {
        current = parsed;
        speaker = segment.participantInfo.identity;
        continue;
      }
    }
```

packages/react/src/hooks/useSessionMessages.ts
thomasyuill-livekit, line 1: I thought this was updated in https://github.com/livekit/components-js/pull/1416?

thomasyuill-livekit (approved):
Generally LGTM

just a few questions and suggestions

Also, looks like there's a linting error tripping up the tests

Would love @1egoman to give it a look too

packages/react/src/hooks/useExpression.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  // The expression rides the stream's *closing* trailer, which livekit-client merges into the
+  // streamInfo object we already hold, by mutation and without emitting. Nothing re-renders, so
+  // the mood would otherwise only surface once the next segment starts: a full turn late. Poll
+  // briefly after each change, and only re-render when the expression actually differs.
+  React.useEffect(() => {
+    let ticks = 0;
+    const id = setInterval(() => {
+      setSettled((previous) => {
+        const latest = segments.reduce<string | null>(
```
thomasyuill-livekit: wouldn't it be faster to go backwards and find the first match?

packages/react/src/hooks/useSessionMessages.ts
1egoman: Maybe this pull request needs to be rebased?

packages/react/src/hooks/useExpression.ts
```diff
@@ -0,0 +1,145 @@
+import * as React from 'react';
+import { useTranscriptions, type UseTranscriptionsOptions } from './useTranscriptions';
+import type { TextStreamData } from '@livekit/components-core';
+
+/** Published per transcript segment by Expressive Mode: the segment's leading delivery tag. */
+export const EXPRESSION_ATTRIBUTE = 'lk.expression';
```
1egoman: nitpick: Put this in here instead: https://github.com/livekit/components-js/blob/2da3e59e9854cde26cbeadcf8a5732ea42163bfa/packages/core/src/helper/participant-attributes.ts#L2

All the agent attributes constants are centralized in that enum.

packages/react/src/hooks/useSessionMessages.ts
thomasyuill-livekit: yeah, I'm sorting this out now and the failed test (generating docs)

packages/react/src/hooks/useExpression.ts
```diff
@@ -0,0 +1,145 @@
...
+ */
+export function parseExpression(segment: TextStreamData): UseExpressionReturn | null {
+  const raw = segment.streamInfo.attributes?.[EXPRESSION_ATTRIBUTE];
+  if (!raw) {
+    return null;
+  }
+
+  try {
+    const parsed = JSON.parse(raw) as { expression?: string; mood?: AgentMood };
+    const expression = parsed.expression?.trim() || null;
```
1egoman, line 80: nitpick: You likely want `??` here instead of `||`:


```suggestion
    const expression = parsed.expression?.trim() ?? null;
```

1egoman (commented):
Generally makes sense to me, just noticed a few nitpicks.

A higher level question - is `useExpression` a good name given how general "expression" is as a term (ie, somebody could think this relates to a mathematical expression, etc) ? Maybe `useAgentExpression` could be better?

1egoman (approved):
Oops, meant to approve

PR #1416, opened 2026-08-05
fix(react): keep text stream attributes on transcript messages

theomonnom:


thomasyuill-livekit (approved):
@theomonnom 
I have no context on what this is for because there is no PR description
but I assume this is necessary for expressive mode

1egoman (approved):
Nice catch, thanks!

Issue #1418, opened 2026-08-09
Chat attachments: attachedFiles ordered by download completion (mergeMap), and received files lose their MIME type


vgerasymenko:
### Summary

Two independent bugs make received chat attachments unusable in `<Chat>` / `<ChatEntry>`:

1. **`attachedFiles` is ordered by download-completion, not by send order** — `setupChat` combines the per-attachment futures with `mergeMap`, which emits as each promise settles. Send three files and they render in whatever order they finished downloading (i.e. smallest first).
2. **A received attachment has no MIME type**, because it is reconstructed as `new File(buffer, fileName)` with no `type` argument — yet `ChatEntry` renders attachments only when `file.type.startsWith('image/')`. That condition is therefore **never true for a received file**, so `<ChatEntry>` renders nothing for an incoming image.

Both are in current `main`. (2) makes (1) invisible with the stock components, but any app with its own entry renderer hits the ordering bug immediately — which is how we found it.

### 1. Ordering — `mergeMap` should be `concatMap`

https://github.com/livekit/components-js/blob/main/packages/core/src/components/chat.ts#L122-L136

```ts
return from(attachments.values()).pipe(
  mergeMap((attachment) => from(attachment.promise)),   // <-- emits in COMPLETION order
  scan(
    (acc, attachment) => [...acc, new File(attachment.buffer, attachment.fileName)],
    [] as Array<File>,
  ),
  map((attachedFiles) => ({ chunk, attachedFiles })),
);
```

`attachments` is a `Map` built from `reader.info.attachedStreamIds`, so it is already in the sender's order. `mergeMap` subscribes to all the promises concurrently and emits each as it resolves, so the `scan` accumulates them in **resolution** order. Replacing it with `concatMap` preserves the map's order at no cost (all the promises are already in flight; `concatMap` only sequences the emissions).

**Observed** (LiveKit Cloud, `livekit-client` 2.21.0, `@livekit/components-react` 2.9.23):

| Sent | Rendered |
|---|---|
| `first.png` (large), `second.png`, `third.png` | second, third, first |
| `image1.png` (large), `doc.pdf` (largest), `image2.png` (small) | image2, image1, doc.pdf |

The pattern is exactly ascending file size. A single-file message is of course unaffected, which is probably why this hasn't been reported.

**Repro**

```ts
// sender
await room.localParticipant.sendText('three files', {
  topic: 'lk.chat',
[...]

PR #1419, opened 2026-08-09
fix(core): preserve chat attachment order and MIME type

vgerasymenko:
Fixes the two bugs reported in #1418.

## 1. `attachedFiles` was ordered by download completion, not send order

`attachments` is built from `reader.info.attachedStreamIds`, so it is already in the sender's order — but `mergeMap` subscribes to all the futures concurrently and emits each as it **resolves**, so the `scan` accumulated them smallest-file-first.

```diff
-              mergeMap((attachment) => from(attachment.promise)),
+              concatMap((attachment) => from(attachment.promise)),
```

`concatMap` only sequences the *emissions* — every promise is already in flight, so there is no added latency, just preserved order.

Observed before the change (LiveKit Cloud, `livekit-client` 2.21.0): three images sent large→small arrived as 2nd, 3rd, 1st; `image1 / pdf / image2` with the pdf largest arrived as image2, image1, pdf.

## 2. Received files had no MIME type

`new File(buffer, fileName)` leaves `File.type` as `''`. Since `ChatEntry` renders attachments behind `file.type.startsWith('image/')`, that condition could **never** be true for a received attachment — so an incoming image rendered as nothing.

The byte stream already exposes `reader.info.mimeType` (non-optional on `BaseStreamInfo`), so it is now carried on the attachment future and passed through:

```diff
       attachment.resolve?.({
         fileName: reader.info.name,
+        mimeType: reader.info.mimeType,
         buffer: bufferList,
       });
```
```diff
-                (acc, attachment) => [...acc, new File(attachment.buffer, attachment.fileName)],
+                (acc, attachment) => [
+                  ...acc,
+                  new File(attachment.buffer, attachment.fileName, { type: attachment.mimeType }),
+                ],
```

A sender's own local echo *does* carry a real type (it is the `File` from the OS picker), which is why both bugs look fine when testing by sending to yourself in one tab.

## Notes

- Scope kept minimal: 15 lines in one file plus a changeset. The outer `mergeMap` (over the text chunk) is unrelated and left alone, so both operators are imported.
- I deliberately did **not** touch `ChatEntry`'s image-only condition. With this fix it works as originally intended; whether to also render non-image attachments (today they are silently dropped) felt like a product decision for maintainers rather than part of a bug fix. Happy to add it if you'd like.
- Changeset included as `patch` for `@livekit/components-core`.

Found while building a multi-participant app where a human operator and an AI agent both exchange chat attachments with a visitor — happy to test a fix against that setup if useful.

### DIFF
--- a/.changeset/chat-attachment-order-and-mimetype.md
+++ b/.changeset/chat-attachment-order-and-mimetype.md
@@ -1,0 +1,5 @@
+---
+'@livekit/components-core': patch
+---
+
+Preserve chat attachment order and MIME type. `setupChat` combined the per-attachment futures with `mergeMap`, which emits as each promise resolves — so `attachedFiles` was ordered by download completion (smallest file first) rather than the order the sender picked them; it now uses `concatMap`, which sequences the emissions of the already-in-flight promises. Received files were also reconstructed as `new File(buffer, fileName)` with no `type`, leaving `File.type` empty — so `ChatEntry`'s `file.type.startsWith('image/')` check could never be true for a received attachment and incoming images rendered as nothing. The byte stream's `mimeType` is now carried through to the `File`.
--- a/packages/core/src/components/chat.ts
+++ b/packages/core/src/components/chat.ts
@@ -112,7 +112,7 @@ export function setupChat(room: Room, options?: ChatOptions) {
       const attachments = new Map(
         (attachedStreamIds ?? []).map((id) => [
           id,
-          new Future<{ fileName: string; buffer: Array<Uint8Array> }, never>(),
+          new Future<{ fileName: string; mimeType: string; buffer: Array<Uint8Array> }, never>(),
         ]),
       );
       streamIdToAttachments.set(id, attachments);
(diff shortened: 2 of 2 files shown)


vgerasymenko:
Pushed 46f58bb to fix the CI failure — that was my mistake.

I added `mimeType` to the module-level `streamIdToAttachments` type but left the per-attachment `new Future<{ fileName; buffer }>()` on its old inline type, so the locally built `Map` wasn't assignable to it. `tsup` succeeded (it strips types) and only the `tsc --declaration --emitDeclarationOnly` step caught it, which is why `test` and `react-size` both failed at the same place — `react-size` builds core first.

`test` and `react-size` are green now. Verified locally before pushing this time, by actually building rather than reasoning about the types:

- `turbo build --filter=@livekit/components-core --force`, including the declaration-emit step
- `pnpm test` in `packages/core` — 98 passing
- `prettier --check` on both changed files

The remaining **Vercel** failure is `Authorization required to deploy`, i.e. a fork-preview permission thing rather than anything in the diff — nothing I can act on from here.

The functional diff is unchanged: `mergeMap` → `concatMap` for ordering, plus threading `reader.info.mimeType` into the `File`.

1egoman (approved):
Good catch, makes sense. Thanks for the contribution!

PR #1425, opened 2026-08-27
fix(core): keep the cached text stream observable across a room disconnect

daniel1014:
### Summary

`setupTextStream` caches one observable per `room:topic` in a global `Map`, registers the handler in `tap({ subscribe })` and unregisters it in `finalize` (refcounted via `share()`). But the `RoomEvent.Disconnected` listener also deletes the cache entry:

```ts
room.on(RoomEvent.Disconnected, () => {
  getObservableCache().delete(cacheKey);   // <-- leftover
  textStreams = [];
  textStreamsSubject.next([]);
});
```

Deleting the entry does **not** invalidate consumers that already hold the observable. So on a reused `Room` instance:

1. Consumer A subscribes while connected — handler registered.
2. The room disconnects. A unsubscribes (`useTextStream` passes `undefined` when disconnected), the handler is unregistered, and the cache entry is dropped — but A still holds observable #1.
3. Consumer B mounts, or an existing consumer changes its `topic` (`useMemo(..., [room, topic])` rebuilds without a remount), while disconnected → cache miss → observable #2 for the same topic.
4. The room reconnects. Both observables' `tap({ subscribe })` fire → `registerTextStreamHandler` is called twice → livekit-client throws:

```
DataStreamError: A text stream handler for topic "lk.transcription" has already been set.
```

The second subscription dies permanently: that topic never delivers text again for the life of the page, while everything else on the room (tracks, other topics) keeps working — which makes it look like a UI bug rather than a stream-handler collision.

This is a leftover from #1188 ("don't unregister stream handler on disconnect"), whose stated goal is exactly this scenario — *"This ensures that a room instance can be reused and the transcription handler stays registered even after a disconnect."* That PR correctly introduced the refcount pattern, but kept the `getObservableCache().delete(cacheKey)` line from the pre-refactor version (where the subject was also completed, so eviction was correct). With the subject now long-lived, the eviction is what breaks reuse.

### Fix

Per @lukasIO's review — key the cache on the `Room` instance itself instead of evicting on disconnect:

```ts
const observableCache = new WeakMap<Room, Map<string, Observable<TextStreamData[]>>>();
```

The entries for a room are collected with the room, so nothing is retained indefinitely, and a reused room keeps its observables across any number of connect/disconnect cycles.

That makes the `RoomEvent.Disconnected` listener unnecessary and it is removed entirely:

- **Cache eviction** is now handled by the `WeakMap`.
- **Buffer reset** moves into `tap({ subscribe })`. `share()` resets on refcount zero, so that callback runs once per subscription window — the first subscriber, and again after every reconnect — which is exactly where a fresh buffer belongs. React consumers already reset their own state on disconnect: `useTextStream` swaps the observable for `undefined`, and `useObservableState` resets to `startWith` whenever the observable identity changes, so dropping the `next([])` emission is not observable to them.

Removing the listener also fixes a leak introduced by #1188, which changed `room.once` → `room.on` without a matching `off`: every cache-key rebuild added another permanent `Disconnected` listener.

This also deletes the `roomInstanceMap` / `nextRoomId` string-key machinery, which only existed to synthesise a key the `WeakMap` now provides directly.

### Repro

Reproduced in an app that mounts `useTextStream` consumers under a `<LiveKitRoom connect={...}>` toggle: connect → disconnect → change a consumer's topic (or mount a new one) → reconnect. The regression tests are the minimal version of that, using a fake `Room` that mirrors livekit-client's one-handler-per-topic contract.

### Test plan

New `packages/core/src/components/textStream.test.ts`:

- `reuses one observable per room and topic across disconnect/reconnect` — red on `main`, and it fails with the production error rather than a bare identity mismatch:
  ```
  AssertionError: expected "vi.fn()" to be called 1 times, but got 2 times
  Unhandled Errors: A text stream handler for topic "lk.transcription" has already been set.
  ```
- `starts each subscription window with an empty buffer` — also red on `main` (`expected [ 2 ] to deeply equal [ 1 ]`); guards the buffer-reset behaviour that moved into `tap({ subscribe })`.
- `caches per room instance, so a second room gets its own observable` — guards the new `WeakMap` keying.

Baseline on `main` (fake room unchanged): `Tests 2 failed | 1 passed (3)`. With the fix: `Tests 3 passed (3)`.

Suites:

- `packages/core`: `pnpm test` → `Test Files 10 passed (10)` / `Tests 101 passed (101)`; `tsc --noEmit` exit 0; `pnpm lint` reports only the 14 pre-existing warnings, none in the touched files.
- `packages/react`: `pnpm test` → `Test Files 4 passed (4)` / `Tests 14 passed (14)`.

Changeset included (`patch` on `@livekit/components-core`).

### DIFF
--- a/.changeset/olive-jokes-shave.md
+++ b/.changeset/olive-jokes-shave.md
@@ -2,4 +2,4 @@
 "@livekit/components-core": patch
 ---
 
-fix(core): keep the cached text stream observable across a room disconnect
+fix(core): key the text stream observable cache on the room instance so it survives a disconnect
--- a/.changeset/olive-jokes-shave.md
+++ b/.changeset/olive-jokes-shave.md
@@ -1,0 +1,5 @@
+---
+"@livekit/components-core": patch
+---
+
+fix(core): keep the cached text stream observable across a room disconnect
--- a/packages/core/src/components/textStream.test.ts
+++ b/packages/core/src/components/textStream.test.ts
@@ -1,0 +1,95 @@
+import { RoomEvent, type Room } from 'livekit-client';
+import { describe, expect, it, vi } from 'vitest';
+import { setupTextStream } from './textStream';
+
+/**
+ * Minimal stand-in for `Room` that mirrors livekit-client's one-handler-per-topic
+ * contract: registering the same topic twice throws.
+ */
+function createFakeRoom() {
+  const handlers = new Map<string, (reader: unknown, participantInfo: unknown) => Promise<void>>();
+  const disconnectListeners: Array<() => void> = [];
+
+  const registerTextStreamHandler = vi.fn(
+    (topic: string, handler: (reader: unknown, participantInfo: unknown) => Promise<void>) => {
+      if (handlers.has(topic)) {
+        throw new Error(`A text stream handler for topic "${topic}" has already been set.`);
+      }
+      handlers.set(topic, handler);
+    },
+  );
+  const unregisterTextStreamHandler = vi.fn((topic: string) => {
+    handlers.delete(topic);
+  });
+
+  const room = {
+    registerTextStreamHandler,
+    unregisterTextStreamHandler,
+    on: (event: RoomEvent, listener: () => void) => {
+      if (event === RoomEvent.Disconnected) disconnectListeners.push(listener);
+      return room;
+    },
+  } as unknown as Room;
+
+  return {
+    room,
+    registerTextStreamHandler,
+    unregisterTextStreamHandler,
+    disconnect: () => disconnectListeners.forEach((listener) => listener()),
+    push: (topic: string, reader: unknown) => handlers.get(topic)?.(reader, { identity: 'agent' }),
+  };
+}
+
+function fakeReader(id: string, chunks: string[]) {
+  return {
+    info: { id, attributes: {} },
+    async *[Symbol.asyncIterator]() {
+      for (const chunk of chunks) yield chunk;
+    },
+  };
+}
+
+const TOPIC = 'lk.transcription';
+
+describe('setupTextStream', () => {
+  it('keeps one observable per topic across a disconnect, so a later subscriber shares the handler', () => {
+    const { room, registerTextStreamHandler, disconnect } = createFakeRoom();
+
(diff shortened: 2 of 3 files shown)


daniel1014:
> [CLA assistant check](https://cla-assistant.io/livekit/components-js?pullRequest=1425) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/livekit/components-js?pullRequest=1425) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/livekit/components-js?pullRequest=1425) it.

Btw I just tried this link to sign the required CLA but the link appears broke blocking me to sign anything on it. It will be really appreciated to have this PR being reviewed and merged as it's a blocking issue around livekit cache (see the details above) requiring manual workaround patch.

lukasIO:
Thanks for the PR! 
Confirmed it's a bug.

However the fix as you suggest would keep the cache entries around indefinitely. 

I'd suggest instead to use a WeakMap keyed on the Room instances like 
```ts
const cache: WeakMap<Room, Map<string, Observable<TextStreamData[]>>>
```

so that the entries can get dropped if the Room instance isn't used anymore. This also allows us to drop the RoomEvent.Disconnected handler entirely (buffer reset can move into `tap({ subscribe })`. 

Let me know if you want to tackle that yourself or if I should create a PR with the fix.

lukasIO:
> Btw I just tried this link to sign the required CLA but the link appears broke 

This is unexpected, do you get an error message that I can use to investigate what might be wrong?

daniel1014:
@lukasIO Thanks! I went with the WeakMap keyed on `Room` and dropped the `Disconnected` listener entirely as you suggested; 
the buffer reset now lives in `tap({ subscribe })`, which `share()` runs once per subscription window. That also let me delete the `roomInstanceMap`/`nextRoomId` key machinery. Both regression tests are red on `main` (the first one fails with the real `A text stream handler ... has already been set` error). Pushed in d704333.

Re the CLA: it shows as signed now — the badge was stale at the time. Sorry for the noise on that.

PR #1426, opened 2026-08-28
fix(core): keep abnormally ended text streams from rethrowing globally

anzemur:
## What

When a participant disconnects while one of its text streams is still open, `livekit-client` errors the open stream's controller with a `DataStreamError` (`AbnormalEnd`) — by design, in `validateParticipantHasNoActiveDataStreams`.

`setupTextStream` reads each incoming stream with `from(reader).pipe(scan(...)).subscribe(next)` — **no error handler**. RxJS then reports the unhandled observable error by rethrowing it globally from a timer (`reportUnhandledError` → `setTimeout(() => { throw err })`), so every abnormally ended transcription stream surfaces as an uncaught `DataStreamError: Participant agent-… unexpectedly disconnected in the middle of sending data` that no application code can catch.

For voice-agent apps this fires whenever the agent leaves the room mid-segment (interrupted farewell, agent process exit, crash). In our production app it is the #1 error by users impacted (~940 users / week) while being functionally benign — the transcript had already been delivered chunk by chunk and the session is over.

The chat handler's `streamObservable.subscribe({ next })` has the same gap.

## Fix

Give both subscribes an `error` callback: keep the already-accumulated text (every chunk was emitted through `next`; on abnormal end there is nothing left to deliver) and log the error at debug level. For chat, `finalize` already cleans up attachment state on error.

## Tests

New test drives `setupTextStream` with a reader that yields chunks and then throws the abnormal-end error: the accumulated text survives, the error is logged, and nothing rethrows globally (without the fix, the test run fails with the uncaught error). Full `@livekit/components-core` suite passes (99 tests).

## Live verification

Reproduced end to end against a real `livekit-server` (Docker, `--dev`) with real Chromium on both sides (Playwright): a sender participant streams four chunks on `lk.transcription` without closing the writer, then goes away; the receiver page runs `setupTextStream` and records all uncaught errors plus emissions. 8/8 deterministic runs across 2 iterations x {unfixed `0.12.15`, this branch} x {tab hard-killed, clean `room.disconnect()` with the stream open}:

- Unfixed: exactly one uncaught page-level `DataStreamError: Participant … unexpectedly disconnected in the middle of sending data` (surfaced via `window.onerror` / Playwright `pageerror` — no application code can catch it).
- Fixed: zero uncaught errors, and the accumulated text still ends complete (`"Hello from the agent"`), so nothing is lost by handling the error.

The only delta between the two receiver bundles is `@livekit/components-core`; every other dependency is alias-pinned identically. The live test drives `textStream.ts`; the identical handler added in `chat.ts` shares the same mechanism but was not separately driven.

## Update: attachment (byte) streams

Review feedback pointed out the fix did not cover attachment streams: a disconnect mid-attachment left the `for await` rejection unhandled, the attachment's `Future` pending forever (it was typed `Future<…, never>` and could not reject), the message pipeline hanging, and the `streamIdToAttachments` entry leaked.

Addressed by settling the future at the source: the byte stream handler now catches the failed read and rejects the attachment's `Future` (error type widened to `Error`). The rejection propagates through the existing message pipeline, so the subscribe `error` callback records it and `finalize` cleans up the attachment state — no hang, no leak, no unhandled rejection, and the pipeline stays healthy for subsequent messages. The error callback is also reason-aware now: `DataStreamErrorReason.AbnormalEnd` logs at debug (expected disconnect churn) while any other failure logs a visible warning.

New tests cover both: an attachment ending abnormally (no unhandled rejection, message dropped, a later message with a completing attachment still delivers) and a non-disconnect attachment failure hitting the warn path.

## Update 2: early rejections and attachments that never start

Two more gaps from review, both fixed:

- **Rejection before anyone listens**: `concatMap` subscribes to the attachment futures one at a time, so a future rejecting before its turn (a later attachment failing first, or any attachment failing before the first text chunk) surfaced as an unhandled rejection. Each future now pre-attaches a no-op `catch` at creation (as suggested), keeping the rejection handled while `concatMap`'s later subscription still observes it.
- **Attachment byte stream never starts**: if the sender drops before opening the byte stream, no controller exists for livekit-client to error, so nothing could ever settle the future - the pipeline hung and the `streamIdToAttachments` entry was retained. The attachment map now records the sender identity, and a `RoomEvent.ParticipantDisconnected` handler (removed in `destroy`) rejects that sender's pending futures with a `DataStreamError` (`AbnormalEnd`).

On the behavior question: the message is **dropped, terminally** rather than delivered text-only - consistent with a mid-transfer attachment failure, so partial failure behaves uniformly (error callback logs, `finalize` cleans up, later messages unaffected). Happy to switch to text-only delivery if that's preferred.

Tests added for both: a later attachment rejecting while the first is still in flight (no unhandled rejection), and a never-starting attachment settled by the disconnect (terminal drop, pipeline healthy for the next message).

### DIFF
--- a/packages/core/src/components/chat.ts
+++ b/packages/core/src/components/chat.ts
@@ -162,6 +162,12 @@ export function setupChat(room: Room, options?: ChatOptions) {
       );
       streamObservable.subscribe({
         next: (value) => messageSubject.next(value),
+        error: (error) => {
+          // Keep an abnormally ended stream (e.g. the sending participant
+          // disconnected mid-stream) from rethrowing globally as an uncaught
+          // exception; `finalize` has already cleaned up its attachment state.
+          log.debug('chat text stream ended abnormally', error);
+        },
       });
     });
     // NOTE: Attachment byte streams are guaranteed to arrive after their parent text stream
--- a/packages/core/src/components/chat.ts
+++ b/packages/core/src/components/chat.ts
@@ -1,6 +1,6 @@
 /* eslint-disable camelcase */
 import type { Room, SendTextOptions } from 'livekit-client';
-import { compareVersions, RoomEvent } from 'livekit-client';
+import { compareVersions, DataStreamError, DataStreamErrorReason, RoomEvent } from 'livekit-client';
 import {
   BehaviorSubject,
   Subject,
@@ -67,7 +67,7 @@ const streamIdToAttachments = new Map<
         mimeType: string;
         buffer: Array<Uint8Array>;
       },
-      never
+      Error
     >
   >
 >();
@@ -112,7 +112,7 @@ export function setupChat(room: Room, options?: ChatOptions) {
       const attachments = new Map(
         (attachedStreamIds ?? []).map((id) => [
           id,
-          new Future<{ fileName: string; mimeType: string; buffer: Array<Uint8Array> }, never>(),
+          new Future<{ fileName: string; mimeType: string; buffer: Array<Uint8Array> }, Error>(),
         ]),
       );
       streamIdToAttachments.set(id, attachments);
@@ -163,10 +163,18 @@ export function setupChat(room: Room, options?: ChatOptions) {
       streamObservable.subscribe({
         next: (value) => messageSubject.next(value),
         error: (error) => {
-          // Keep an abnormally ended stream (e.g. the sending participant
-          // disconnected mid-stream) from rethrowing globally as an uncaught
+          // Keep a failed message (text stream errored, or an attachment
+          // future rejected) from rethrowing globally as an uncaught
           // exception; `finalize` has already cleaned up its attachment state.
-          log.debug('chat text stream ended abnormally', error);
+          // A disconnect mid-stream is expected churn; anything else deserves
+          // a visible warning.
+          const abnormalEnd =
+            error instanceof DataStreamError && error.reason === DataStreamErrorReason.AbnormalEnd;
+          if (abnormalEnd) {
+            log.debug('chat message stream ended abnormally', error);
+          } else {
+            log.warn('chat message stream failed', error);
+          }
         },
       });
     });
@@ -183,8 +191,20 @@ export function setupChat(room: Room, options?: ChatOptions) {
       const streamId = foundStreamAttachmentPair[0];
 
       const bufferList = [];
-      for await (const buffer of reader) {
-        bufferList.push(buffer);
+      try {
+        for await (const buffer of reader) {
+          bufferList.push(buffer);
+        }
+      } catch (error) {
+        // Settle the attachment future so the message pipeline errors instead
+        // of hanging forever - its error callback logs and `finalize` cleans up
+        // the attachment state. Without this the rejection is unhandled and the
+        // pending future leaks its `streamIdToAttachments` entry.
+        streamIdToAttachments
+          .get(streamId)
+          ?.get(attachmentStreamId)
+          ?.reject?.(error instanceof Error ? error : new Error(String(error)));
+        return;
       }
 
       const attachment = streamIdToAttachments.get(streamId)?.get(attachmentStreamId);
--- a/packages/core/src/components/chat.ts
+++ b/packages/core/src/components/chat.ts
@@ -1,5 +1,5 @@
 /* eslint-disable camelcase */
-import type { Room, SendTextOptions } from 'livekit-client';
+import type { Participant, Room, SendTextOptions } from 'livekit-client';
 import { compareVersions, DataStreamError, DataStreamErrorReason, RoomEvent } from 'livekit-client';
 import {
   BehaviorSubject,
@@ -59,17 +59,20 @@ export type ChatOptions = {
 const topicSubjectMap: WeakMap<Room, Map<string, Subject<ReceivedChatMessage>>> = new WeakMap();
 const streamIdToAttachments = new Map<
   string /* stream id */,
-  Map<
-    string /* attachment id */,
-    Future<
-      {
-        fileName: string;
-        mimeType: string;
-        buffer: Array<Uint8Array>;
-      },
-      Error
-    >
-  >
+  {
+    senderIdentity: string;
+    attachments: Map<
+      string /* attachment id */,
+      Future<
+        {
+          fileName: string;
+          mimeType: string;
+          buffer: Array<Uint8Array>;
+        },
+        Error
+      >
+    >;
+  }
 >();
 
 function isIgnorableChatMessage(msg: ReceivedChatMessage | LegacyReceivedChatMessage) {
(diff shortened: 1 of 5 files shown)


1egoman:
@anzemur Thanks for the contribution! There's a few issues though which I think would need to be addressed prior to being merged:

#### 1. The change does not correct the problem for data stream attachments.
For context, the way data streams works, the sender transmits each attachment as a separate data stream forming a "tree". The `attachedStreamIds` field links each attachment stream to its parent text stream, and `setupChat` makes one `Future` for each attachment. The chat message pipeline waits for all of these Future objects before resolving.

If the sender disconnects during an attachment stream, the `for await` loop in the byte stream handler throws an error. The byte stream handler is an async function, and `livekit-client` does not catch the rejection. Because of this, the error becomes an unhandled rejection, and the new error callback does not receive this error, because the error is not in the text stream.

This means the attachment `Future` does not resolve or reject, it stays pending indefinitely. This means the pipeline does not complete, finalize does not run, and the key stays in `streamIdToAttachments` permanently. The user also does not receive the chat message.

So, I don't think a top-level catch is sufficient. Since no error bubbles up to the top level, you first need to settle the attachment `Future`, then, the top-level handler can record the error in one place.

#### 2. Errors in the text stream are safe, but errors in an attachment stream are not

If the text stream fails during a read, finalize runs for all error types and deletes all state in any persistent maps. This is correct today.

However, if an attachment stream fails during a read, the state stays. This is true for all error types, not only for a disconnection. So I think you would need to make this case disconenction error specific, which you can do by reading the `DataStreamErrorReason` enum on the error.

General thoughts on these issues? I think as part of this change it might be worthwhile to introduce at least one more (if not more) test(s) with a data stream that has an attachment and validate that errors propegate properly in the above situations.

packages/core/src/components/chat.ts
```diff
@@ -112,7 +112,7 @@ export function setupChat(room: Room, options?: ChatOptions) {
       const attachments = new Map(
         (attachedStreamIds ?? []).map((id) => [
           id,
-          new Future<{ fileName: string; mimeType: string; buffer: Array<Uint8Array> }, never>(),
+          new Future<{ fileName: string; mimeType: string; buffer: Array<Uint8Array> }, Error>(),
         ]),
       );
```
1egoman, line 126: I suspect here you want something like the below, otherwise if you have multiple attachments which are included on chat message, there's a narrow period of time in the `concatMap` in `chat.ts` line 134 where if a supposed attachment index > 0 were to reject, it would get raised as an unhandled rejection. I think to work around this, you'd want something like the below here:

```suggestion
      const attachments = new Map(
        (attachedStreamIds ?? []).map((id) => {
          const future = new Future<{ fileName: string; mimeType: string; buffer: Array<Uint8Array> }, Error>();
          future.promise.catch(() => {}); // Ignore emitting `unhandledRejection` if the promise rejects before the attachments `concatMap` switches to this promise
          return [id, future];
        }),
      );
```

Example test which I think should exercise this case:
```ts
const byteReader = (id: string, body: () => AsyncGenerator<Uint8Array>): ByteReader => ({
  info: { id, name: `${id}.bin`, mimeType: 'application/octet-stream' },
  [Symbol.asyncIterator]: body,
});

it('does not reject an attachment future before anything subscribes to it (multiple attachments)', async () => {
  const rejections: Array<string> = [];
  const onUnhandled = () => rejections.push('unhandledRejection');
  process.on('unhandledRejection', onUnhandled);
  vi.spyOn(log, 'debug').mockImplementation(() => {});
  vi.spyOn(log, 'warn').mockImplementation(() => {});
  try {
    const { room, textHandlers, byteHandlers } = makeRoom();
    const chat = setupChat(room);
    const subscription = chat.messageObservable.subscribe(() => {});

    const textHandler = textHandlers.get('lk.chat')!;
    const byteHandler = byteHandlers.get('lk.chat')!;

    await textHandler(textReader('msg-1', 'two attachments', ['att-1', 'att-2']), {
      identity: 'sender',
    });
    await settle();

    // `concatMap` subscribes to the attachment futures one at a time, so nothing is
    // listening to att-2 while att-1 is still in flight. Rejecting it here surfaces
    // globally as an unhandled rejection until att-1 settles.
    await byteHandler(
      byteReader('att-2', async function* () {
        throw new DataStreamError(
          'Participant sender unexpectedly disconnected in the middle of sending data',
          DataStreamErrorReason.AbnormalEnd,
        );
      }),
    );
    await settle();

    expect(rejections).toEqual([]);

    subscription.unsubscribe();
  } finally {
    process.off('unhandledRejection', onUnhandled);
    vi.restoreAllMocks();
  }
});
```

packages/core/src/components/chat.ts
1egoman, line 135: Another case that I don't think your current solution handles properly - a byte stream fails before the text stream yields its first chunk which results in an unhandled rejection.

The attachment observable is only constructed inside `mergeMap` here, which doesn't fire until a text chunk arrives. Similarly to my other comment, if the downstream attachment is rejected before that and nobody upstream is listening with a `.catch()1, then you will get an unhandledRejection.


I think a test which should assert this:

```ts
const byteReader = (id: string, body: () => AsyncGenerator<Uint8Array>): ByteReader => ({
  info: { id, name: `${id}.bin`, mimeType: 'application/octet-stream' },
  [Symbol.asyncIterator]: body,
});

it('settles the message when an attachment byte stream never arrives at all', async () => {
  vi.spyOn(log, 'debug').mockImplementation(() => {});
  vi.spyOn(log, 'warn').mockImplementation(() => {});
  try {
    const { room, textHandlers } = makeRoom();
    const chat = setupChat(room);
    const emissions: ReceivedChatMessage[][] = [];
    const subscription = chat.messageObservable.subscribe((me
      emissions.push(messages);
    });

    const textHandler = textHandlers.get('lk.chat')!;

    // The text arrives complete, but the sender drops before
    // byte stream, so its future never settles. Nothing errors, so no error callback
    // can rescue this: the pipeline never completes, `finali
    // `streamIdToAttachments` entry is retained for the lifetime of the page.
    await textHandler(textReader('msg-3', 'text arrived, atta']), {
      identity: 'sender',
    });
    await settle();

    // The received text must not be silently swallowed. (If the preferred behaviour is
    // to drop the message instead, assert that the pipeline
    // that it must reach *some* terminal state rather than hanging forever.)
    expect(emissions.flat().map((message) => message.message)
      'text arrived, attachment never did',
    ]);

    subscription.unsubscribe();
  } finally {
    vi.restoreAllMocks();
  }
});
```

1egoman (commented):
Thanks for the continued progress on this! I think you are on the right track, but there's a few small bugs with with what you have now which I think would need to be addressed prior to being merged.

packages/core/src/components/chat.ts
```diff
@@ -291,13 +326,34 @@ export function setupChat(room: Room, options?: ChatOptions) {
...
+        // terminal state instead of hanging. Settled futures ignore this.
+        attachment.reject?.(
+          new DataStreamError(
+            `Participant ${participant.identity} disconnected before sending all attachments`,
+            DataStreamErrorReason.AbnormalEnd,
+          ),
+        );
+      }
+    }
+  };
```
1egoman, line 346: I am a bit concerned with this approach as re-raising this error drops all context as the original, source error which was raised by the sdk, and that this is the only place which would create a `DataStreamError` outside of the client-sdk-js core, which is fairly unusual.

Admittedly I suspect you've gone deeper than me into this particular failure mode, but I would think there would be _some_ sort of way to filter the existing stream of errors rather than converting all to logs and reraising in some situations. Is there some reason this solution wouldn't work? Just want to make sure I'm not missing something here.

packages/core/src/components/chat.ts
```diff
@@ -291,13 +326,34 @@ export function setupChat(room: Room, options?: ChatOptions) {
     }
   };
 
+  const handleParticipantDisconnected = (participant: Participant) => {
+    for (const { senderIdentity, attachments } of streamIdToAttachments.values()) {
+      if (senderIdentity !== participant.identity) {
+        continue;
+      }
+      for (const attachment of attachments.values()) {
```
1egoman, line 334: (Note - I am not sure if this remains relevant given my previous comment, but as a FYI, if this code sticks around, you need to somehow filter by room here as well. As it is currently, I believe it is possible for a disconnect of a participant with a given identity on one room to cause this code to get hit also for another room which is still open.)

PR #1447, opened 2026-09-19
fix(react): refresh session screen share state

fatihcvs:
When a connected session starts or stops screen sharing without changing its camera, microphone, or connection state, `session.local.screenShareTrack` stays stale. Add `localScreenShare` to the session-state memo dependencies.

The regression tests publish and unpublish a screen-share publication on a real SDK `Room`, with connection preparation mocked. Both cases fail before the fix and pass afterward. Includes a patch changeset.

Validated locally with Node 24.13.0 on Windows:
- `pnpm build:react` (styles, core, React)
- React/core/styles tests: 120 passing
- React lint (existing warnings), `pnpm format:check`
- React `api-check`

Developed with Codex assistance.


1egoman (approved):
I think this makes sense, thanks for the contribution!

fatihcvs:
You're welcome! Thanks for the review.

Issue #524, opened 2023-06-13
Focusing a track from the carousel leads to: Error while running updatePages()


Ocupe:
### Select which package(s) are affected

\@livekit/components-react

### Describe the bug

There should be no error in the console when you pin a participant from the carousel. Dieser Fehler scheint weder die Funktion des Karussells noch den Rest der App zu beeinträchtigen. 

### Reproduction

When one participant is focused, focus another participant from the carousel. 

### Logs

<img width="684" alt="image" src="https://github.com/livekit/components-js/assets/11357413/3156addc-7812-43a3-b120-d2da6473f05e">

### System Info

```shell
System:
    OS: macOS 13.4
    CPU: (10) arm64 Apple M1 Max
    Memory: 2.28 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.2.0 - /opt/homebrew/bin/node
    Yarn: 1.22.19 - ~/Library/pnpm/yarn
    npm: 9.6.6 - /opt/homebrew/bin/npm
  Browsers:
    Brave Browser: 112.1.50.119
    Chrome: 112.0.5615.137
    Firefox: 114.0
    Firefox Developer Edition: 105.0
    Safari: 16.5
```

### Severity

annoyance

lebaudantoine:
same error, when I quit a meeting, I got the same error @Ocupe 
```
Error while running updatePages():  Error: Element not part of the array: Antoine__ug6s_camera_placeholder not in Antoine__ug6s_camera_TR_VCt38mfcudpVDY
    at t1 (278-bac2459ffa7ea51d.js:3:28533)
    at 278-bac2459ffa7ea51d.js:3:29308
    at 278-bac2459ffa7ea51d.js:3:29367
    at Array.forEach (<anonymous>)
    at 278-bac2459ffa7ea51d.js:3:29258
    at Array.forEach (<anonymous>)
    at t2 (278-bac2459ffa7ea51d.js:3:28969)
    at k (278-bac2459ffa7ea51d.js:3:51646)
    at S (278-bac2459ffa7ea51d.js:3:51920)
    at em (278-bac2459ffa7ea51d.js:1:84661)
    ```
    
    I could reproduce it in my app, and your demo app !

Ocupe:
The logic seems OK, but the order of operation seems off. I think we are calling the hook at a weird inbetween state.
@lebaudantoine, does it have any negative impact besides the error logged to the console?

[...]

PR #1448, opened 2026-09-19
fix(core): handle track replacements in visual layout updates

fatihcvs:
`updatePages` only adds new tracks when the total list length increases. If a track is replaced while another is removed (for example, switching the focused participant or replacing an unpublished camera with its placeholder), the swap searches for an item that was never added and throws `Element not part of the array`.

Always add the missing items before updating pages, preserving the existing page order and removal logic. Related to #524.

Validation:

- Four regression cases cover equal-size replacement, replacement while shrinking/growing, and camera-to-placeholder replacement. Three fail on the original code with the reported exception; all 43 sorting tests pass after the fix.
- 122 core/react/styles tests pass. `pnpm build:react`, core lint (14 existing warnings), `pnpm format:check`, and React API check pass.
- A separate check against the built core package covered 16,900 transitions of unique four-item subsets/permutations across page sizes 1–4, verifying exact output membership and unchanged inputs. No live meeting/browser call was used for this check.

Includes a core patch changeset. Developed and verified with OpenAI Codex assistance.


1egoman (approved):
I think this makes sense, thanks for the contribution!

fatihcvs:
You're welcome! Thanks for reviewing the track replacement fix.

PR #1449, opened 2026-09-19
fix(core): initialize chat receivers for each topic

fatihcvs:
Calling setupChat for two different channelTopic values in the same room only installed receivers for the first topic. Repeated calls also installed a disconnect listener per consumer, causing the same topic to be torn down repeatedly.

Initialization now checks the topic within the room map. Each topic owns its message and destruction subjects, while one once-only disconnect listener tears down all topics for that room. It completes subscribers, removes legacy subscriptions and unregisters each text/byte handler once. A subsequent setup installs fresh state and one new listener.

Validation:
- The two original regression tests failed before the initialization fix. The added teardown regression failed before the review fix (three added listeners instead of one).
- Tests cover independent legacy channels, repeated setup, second-topic text streams, observer completion, exactly-once handler removal, repeated disconnects, two setup/disconnect cycles and isolation from another room.
- Real SDK Room events and handler registration are used, with an async-iterator reader test double. No LiveKit server connection was made.
- Current validation: 101 core + 14 React + 6 styles tests passed; core build/declaration generation, core lint (14 existing warnings, no errors), changed-file Prettier and git diff checks passed.

Includes an updated core patch changeset.


packages/core/src/components/chat.ts
1egoman, line 299: Since you are fixing the initialization end, would it be possible for you to also fix the teardown end too so it has the same sematics? I believe today a new listener is added per `setupChat()` call (which happens once per `useChat()` / etc call) so this teardown can potentially run many times.

I think (though I haven't tested this) that something like the below is what should be here instead:
```tsx
const isFirstTopicForRoom = !topicSubjectMap.has(room);
if (isFirstTopicForRoom) {
  room.on(RoomEvent.Disconnected, () => {
    const topics = topicSubjectMap.get(room);
    topicSubjectMap.delete(room);
    topics?.forEach(({ messageSubject: subject, onDestroyObservable: onDestroy }, chatTopic) => {
      onDestroy.next();
      onDestroy.complete();
      subject.complete();
      room.unregisterTextStreamHandler(chatTopic);
      room.unregisterByteStreamHandler(chatTopic);
    });
  });
}
```

And then multiple `setupChat()` calls via `useChat()` or similar should only ever result in one initialization and one teardown per topic.

1egoman (commented):
This generally looks good to me - ideally it would be great if you could also address the teardown side as well, but if you don't have the bandwidth let me know and I can make an issue for this to be addressed later.

1egoman (approved):
Looks good to me, thanks for the contribution!

fatihcvs:
You're welcome! Thanks for reviewing it.

